PR #1, opened 2022-11-30
Scc 3261/stubbed index

charmingduchess:


lib/clients/platform-api/client.js
```diff
@@ -0,0 +1,33 @@
...
+      log_level: 'error'
+    })
+  }
+
+  return clientPromise
+}
+
+module.exports = {
+  init
+}
```
nonword, line 33: I'm not sure about the division between "clients" and "things that aren't clients". If I'm looking for how the app integrates KMS, I'd think twice about whether that's a "client" or not, so I'm not sure placing modules that have client-like responsibilities (i.e. interface external apis) into a `clients` folder helps. It also creates a problem in that it's unclear what to do with functions that leverage client functions, but aren't strictly part of the client (e.g. all of the things in `get-fors.js`). I think subfoldering the `lib` folder is fine, but maybe just `lib/platform-api` (containing `client.js` and `get-fors.js`) for example. (Separately, I like placing convenience functions in `get-fors` but I'm not sure that's the best name. Maybe just `utils` (i.e. `lib/platform-api/utils.js`)?

lib/event-decoder.js
```diff
@@ -0,0 +1,47 @@
+const avro = require('avsc')
+
+const logger = require('./clients/logger')
+const platformApi = require('./platform-api')
```
nonword, line 4: This loads which `platform-api` lib? Isn't this a directory?

nonword (commented):
Looking good. I like how `index.js` is looking. Really helps to see how it's going to come together.

nonword (commented):
A couple comments. See also [this pr with suggestions](https://github.com/NYPL/research-catalog-indexer/pull/7)

lib/platform-api/requests.js
```diff
@@ -0,0 +1,66 @@
...
+  )
+  bibs = bibs.filter((b) => b)
+
+  return bibs
+}
+
+const bibsForHoldings = (holdings) => {
+  return _bibIdentifiersForHoldings(holdings)
+    .map((ids) => bibById(ids.nyplSource, ids.id))
+}
```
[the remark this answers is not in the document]
charmingduchess: bib identifiers for holdings doesnt make any api calls, at least as it was written in the original indexer

lib/event-decoder.js
```diff
@@ -0,0 +1,49 @@
...
+
+async function _getSchema (schemaName) {
+  // schema in cache; just return it as a instant promise
+  if (CACHE[schemaName]) {
+    logger.debug(`Already have ${schemaName} schema`)
+    return Promise.resolve(CACHE[schemaName])
+  }
+
+  // Fetch schema and parse it as an AVSC decoder
+  const { data: schemaObject } = await platformApi.getSchema(schemaName)
```
nonword, line 16: Doesn't this assign the value of the `data` property to a new variable called `schemaObject`? (That is, your `schemaObject` takes on the complete value of response.data) I think this should actually be `const { data: { schemaObject: schemaObject } } = await ...` or `const { data: { schemaObject } } = await ...` for brevity

nonword: But also, looks like `lib/platform-api/requests#getSchema` is returning `response.data`, so no need to destructure the `data` prop

nonword (commented):
Looking good. At this stage, we should be testing it with `sam` as well, so I think you should add a basic sam file like
```
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: 'research-catalog-indexer'

Resources:
  ResearchCatalogIndexer:
    Type: AWS::Serverless::Function
    Properties:
      Handler: index.handler
      Runtime: nodejs18.x
      Timeout: 300
      Environment:
        Variables:
          NYPL_API_BASE_URL: http://qa-platform.nypl.org/api/v0.1/
          NYPL_OAUTH_URL: https://isso.nypl.org/
          NYPL_OAUTH_KEY: AQECAHh7ea2tyZ6phZgT4B9BDKwguhlFtRC6hgt+7HbmeFsrsgAAAGswaQYJKoZIhvcNAQcGoFwwWgIBADBVBgkqhkiG9w0BBwEwHgYJYIZIAWUDBAEuMBEEDNQozUGkaz8WYD2lUAIBEIAo/SzNMA9LowO6gcnTUCcMjBaAU1RH/L3EAS14fjJCUpyZppkuEDUd7w==
          NYPL_OAUTH_SECRET: AQECAHh7ea2tyZ6phZgT4B9BDKwguhlFtRC6hgt+7HbmeFsrsgAAAIcwgYQGCSqGSIb3DQEHBqB3MHUCAQAwcAYJKoZIhvcNAQcBMB4GCWCGSAFlAwQBLjARBAxN3indXvk2ueiE6CwCARCAQ018FdIVXXwfTuKH1vp/ZTfjBinxKTDosMmzyWB9/CtiFtgOu09iiyZEpC3AyGOt8ExywHZoHOZQuLdGGNFgbusmldw=
          LOGLEVEL: debug
```
And try invoking it like `sam local invoke --profile nypl-digital-dev -t sam.qa.yml -e test/sample-events/b10128427.json` (copying sample events over from DHI) as that kind of testing will uncover the kinds of runtime issues I noted

lib/event-decoder.js
charmingduchess: yes! good catch on both counts

lib/stubzzz.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+  return filter
+}
+
+const asyncNoop = async (x) => Promise.resolve(x)
+class EsBib {
+  constructor (bib) {
+    Object.keys(bib).forEach((key) => {
+      this[key] = bib[key]
+    })
```
nonword, line 13: This is just a stub, but it looks like this pattern is messing with the `toJson` process run on this stubbed data later. I see errors when running this via `sam`. The actual implementation of `EsBib` should store the SierraBib model as an instance variable instead of cloning all of `bib`'s properties into itself. e.g.
```
class EsBib {
  constructor (bib) {
    this.bib = bib 
  }
  ...
}
```

lib/kms.js
```diff
@@ -0,0 +1,17 @@
+const { KMSClient, DecryptCommand } = require('@aws-sdk/client-kms')
+const logger = require('./logger')
+
+const decrypt = async (encrypted) => {
+  try {
+    const client = new KMSClient({
+      region: process.env.AWS_REGION || 'us-east-1'
+    })
+    const command = new DecryptCommand(encrypted)
```
nonword, line 9: I think the constructor to `DecryptCommand` [takes something different](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-kms/classes/decryptcommand.html) than just the encrypted string? e.g. 
```
 new DecryptCommand({ CiphertextBlob: Buffer.from(encrypted, 'base64') })
```
You'll see errors if you run `sam invoke..`

danamansana (dismissed):

nonword (approved):

nonword (approved):

PR #9, opened 2022-12-07
Add initial EsBib model and tests

danamansana:
- Adds initial EsBib model and EsBase class
- adds initial tests for EsBib
- adds some utilities for use in EsBib
- adds initial bib mappings and some common mapping utilities
- still looking for examples for a couple of EsBib methods, also missing tests for utility functions


lib/utils/utils.js
```diff
@@ -0,0 +1,36 @@
...
+      return 'b'
+    case 'recap-pul':
+      return 'pb'
+    case 'recap-cul':
+      return 'cb'
+    case 'recap-hl':
+      return 'hb'
+    default:
+      return ''
+  }
```
nonword, line 13: We should really have this be driven from [the central "source mapping" file in NYPL-core](https://github.com/NYPL/nypl-core/blob/master/mappings/recap-discovery/nypl-source-mapping.json). There's [a util we have used](https://github.com/NYPL-discovery/discovery-store-models/blob/40acac46a231b9eac185e7670df7eea5480378c6/lib/nypl-source-mapper.js) for interfacing that data when needing to convert prefixed identifiers into their component keys (id, nyplSource). You want to do the reverse, but there will likely be a need for converting in both directions, so I'd copy that NyplSourceMapper lib into this repo and amend it to your needs. (Note that the `syncRequest` call isn't going to work in Node 18, so you'll need to make it async.) What you need to do here is convert an nyplSource into a prefix, which could be done by loading the [above source mapping json](https://github.com/NYPL/nypl-core/blob/master/mappings/recap-discovery/nypl-source-mapping.json) and then returning `mapping[nyplSource].bibPrefix`. We don't add partners often, but it sure helps to have it centralized

lib/utils/nypl-source-mapper.js
```diff
@@ -0,0 +1,58 @@
...
+
+  /** Given an nypl source (such as sierra-nypl or recap-pul) returns the matching prefix
+   */
+  async prefix (source) {
+    const nyplSourceMapping = await this.nyplSourceMapping()
+    return nyplSourceMapping[source].bibPrefix
+  }
+}
+
+module.exports = NyplSourceMapper
```
nonword, line 58: Looks good. Thanks for taking on making this module async

lib/es-models/base.js
```diff
@@ -0,0 +1,7 @@
+class EsBase {
+  updatedAt () {
```
charmingduchess, line 2: I included an updatedAt function in the indexing function. Not sure if it's better off here or not. I think there will have to be some work to reconcile any duplicated or convoluted logic some time down the line. What do you think, @nonword ?

lib/es-models/bib.js
```diff
@@ -0,0 +1,73 @@
...
+    const mappings = BibMappings.get('contributorLiteral', this.bib)
+    return parallelValues(mappings, this.bib)
+  }
+
+  numItems () {
+    return 0 // TODO
+  }
+
+  type () {
+    return 'nypl:Item'
```
charmingduchess, line 65: should this be nypl:bib given that it's in the bib model?

danamansana: Yeah I was confused about this, too but it is correct

charmingduchess: can you insert a comment that calls that out? for posterity!

lib/es-models/base.js
nonword: We want an `updatedAt` property on the bib, but not on each item or holding (because the current meaning of this `updatedAt` is "what time was this indexed?", so all `updatedAt` values would be the same if we added it to items and holdings). So I think it might need to go in `EsBib` only. Otherwise, the indexing will fail when it attempts to add `updatedAt` properties to items and holdings.

lib/es-models/bib.js
danamansana: Sure! Added a small comment for clarity

lib/es-models/base.js
danamansana: done!

nonword (dismissed):

nonword (commented):
Looks like we need to resolve some upstream package-lock issues. After that, looks good.

PR #13, opened 2022-12-28
Add prefiltering methods and tests

danamansana:
- Adds prefiltering of bibs and items
- Adds `fixed` method in Sierra Base class to fetch fixed fields following the same method as currently used in DHI
- Pulls in `nypl-core-objects` package (needed to distinguish between research and branch locations)
- Adds tests


lib/utils.js
```diff
@@ -1,7 +1,75 @@
...
   return resp && resp.data
 }
 
+/**
+* Given an array of items, resolves a new array of items containing only
+* those items that *may* be Research based on Item Type
+*/
+const prefilterItems = async function (items) {
+  const filteredItems = items.filter((item) => {
+    if (item.nyplSource !== 'sierra-nypl') return true
```
charmingduchess, line 16: can you put comments describing whats happening for each of these checks? ie like 
`// if its not a sierra item, its a partner item, hence research `
(like you have for the bibprefilter)

charmingduchess (dismissed):
looks good!

nonword (dismissed):

PR #19, opened 2023-01-30
Scc 3387/fix parallels

charmingduchess:
this work updates the SierraBase.varField() function to attach parallels to primary values and include orphan parallels
this is a multi step process, including:
- parallelsCache, a instance variable object with key value pairs relating subfield 6 to the raw parallel field that contains that subfield 6. created in the constructor function
- _addParallels(), which takes a marc value, generates 2 arrays of the corresponding fields, one for primary values and one for primary values. the parallel fields are found by searching through the parallels cache for keys that include the marc value. because we need to account for parallels with no primaries and primaries with no parallels, we have to loop through both arrays separately. matched parallels are nullified in an internal lookup to facilitate this process. the function returns an array of mostly raw fields, with an added property of parallel (when appropriate)
- _prettifyRawVarFields(), the new name for _extractValuesfromVarfields, takes the array returned from _addParallels and returns a reformatted varField that we are familiar with, including the parallel field when appropriate

other updates include using the full script name instead of the script code for parallel values, as well as updating tests to match new data structures.


lib/sierra-models/base.js
```diff
@@ -20,110 +21,95 @@ class SierraBase {
 
   fieldTag (field, subfields, opts) {
     const varFields = this._varFieldByFieldTag(field)
-    return this._extractValuesFromVarFields(varFields, subfields, opts)
+    return this._prettifyRawVarFields(varFields, subfields, opts)
```
nonword, line 24: I'm not sure "prettify" is the right term here, but I hear you that "_extractValuesFromVarFields" also doesn't really capture it. "Prettification" tends to mean formatting. What `_prettifyRawVarFields` is doing is transforming raw varfields into VarFieldMatch objects.

lib/sierra-models/base.js
```diff
@@ -20,110 +21,95 @@ class SierraBase {
...
+      return field
+    })
+
+    const orphanParallels = parallelRawFields.map((field) => {
+      // check if parallel remains unmatched in lookup table.
+      //   if it is, it will be returned {parallel: {value:'meep', subfieldMap: {'a': 'meep'}}}
+      lookup(field)
+      if (field.parallel) {
+        return field
+      } else { return null }
```
nonword, line 93: Something in here isn't right. An orphaned parallel is rendered as
```
{
  value: "parallel value",
  ...
  parallel: {
    value: "parallel value",
    ...
  }
}
```
See [this pr](https://github.com/NYPL/research-catalog-indexer/pull/20) with some suggestions to address this bug and clarify a few things

nonword (commented):
Looking good. I think there may still be an issue with orphaned parallels. I think there are also legacy parts of the code that benefit from some refactoring. See comments.

nonword (approved):

PR #20, opened 2023-01-31
Refactoring and fixing orphaned parallel extraction

nonword:
Some proposals to:
 - fix orphaned parallel extraction, which wasn't handled properly before (parallel values were taking the place of primary values)
 - refactor the "prettify" function for clarity
 - add tests to cover some contrived edge cases with orphaned parallels

### DIFF
--- a/lib/sierra-models/base.js
+++ b/lib/sierra-models/base.js
@@ -46,7 +46,7 @@ class SierraBase {
   // This function takes a marc tag, and returns an array of primary values with
   //   relevant parallel values attached and or orphaned parallels relevant to
   //   given marc tag
-  _addParallels (marc) {
+  _fieldsWithParallelsAndOrPrimaries (marc) {
     // _varFieldByMarcTag returns array of raw marc fields filtered by marc tag:
     //  [{fieldTag: 'a', marcTag: '123', content: 'the thing'}]
     const primaryRawFields = this._varFieldByMarcTag(marc)
@@ -64,7 +64,7 @@ class SierraBase {
       return { ...lookup, [number]: rawField }
     }, {})
 
-    const lookup = (field) => {
+    const attachParallelToPrimary = (field) => {
       const subfieldSix = this._parseSubfield6(field)
       if (subfieldSix) {
         const { number: subfieldSixSuffix } = subfieldSix
@@ -73,25 +73,19 @@ class SierraBase {
           //    add parallel to the primary
           field.parallel = { ...parallelsLookup[subfieldSixSuffix] }
           //    set value to null, so we know that parallel has been matched
-          parallelsLookup[subfieldSixSuffix] = null
+          delete parallelsLookup[subfieldSixSuffix]
         }
       }
+      return field
     }
 
-    const primariesWithAndWithoutMatches = primaryRawFields.map((field, i) => {
-      // attach relevant parallels
-      lookup(field)
-      return field
-    })
+    const primariesWithAndWithoutMatches = primaryRawFields
+      .map(attachParallelToPrimary)
 
-    const orphanParallels = parallelRawFields.map((field) => {
-      // check if parallel remains unmatched in lookup table.
-      //   if it is, it will be returned {parallel: {value:'meep', subfieldMap: {'a': 'meep'}}}
-      lookup(field)
-      if (field.parallel) {
-        return field
-      } else { return null }
-    }).filter((p) => p)
+    const orphanParallels = Object.values(parallelsLookup)
+      .map((parallel) => ({ parallel }))
+
+    // Join primaries (with attached parallels) to orphaned parallels
     return primariesWithAndWithoutMatches.concat(orphanParallels)
   }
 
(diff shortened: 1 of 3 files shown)


lib/sierra-models/base.js
```diff
@@ -46,7 +46,7 @@ class SierraBase {
   // This function takes a marc tag, and returns an array of primary values with
   //   relevant parallel values attached and or orphaned parallels relevant to
   //   given marc tag
-  _addParallels (marc) {
```
nonword, line 49: This function appears to do much more than adding a parallel, so proposing name change

lib/sierra-models/base.js
```diff
@@ -64,7 +64,7 @@ class SierraBase {
       return { ...lookup, [number]: rawField }
     }, {})
 
-    const lookup = (field) => {
```
nonword, line 67: This is a misleading name for a function that attaches a raw parallel field to the given primary field.

lib/sierra-models/base.js
```diff
@@ -73,25 +73,19 @@ class SierraBase {
           //    add parallel to the primary
           field.parallel = { ...parallelsLookup[subfieldSixSuffix] }
           //    set value to null, so we know that parallel has been matched
-          parallelsLookup[subfieldSixSuffix] = null
```
nonword, line 76: Proposing just removing the entry from the hash entirely so we don't have to filter on nulls later

lib/sierra-models/base.js
```diff
@@ -73,25 +73,19 @@ class SierraBase {
...
         }
       }
+      return field
     }
 
-    const primariesWithAndWithoutMatches = primaryRawFields.map((field, i) => {
-      // attach relevant parallels
-      lookup(field)
-      return field
-    })
```
nonword, line 85: Again, confusing that something called `lookup` mutates the element. Feels like a `map` using a mutating function

lib/sierra-models/base.js
```diff
@@ -136,7 +131,7 @@ class SierraBase {
 
   // returns array of objects. value is concatenation of contents of
   //   specified subfields - {value: 'val parallel link', subfieldMap: {a:'val', 6:'parallel link'}}
-  _prettifyRawVarFields (rawFields, subfields, opts) {
```
nonword, line 139: The legacy choice of "subfields" here creates naming issues later, so proposing a name that conveys this is a config argument indicating "which subfields" to use

lib/sierra-models/base.js
```diff
@@ -147,42 +142,47 @@ class SierraBase {
...
+          varFieldMatchReturn.value = field.content
         }
 
-        // If asked to match certain subfields, return only those:
-        if (subfields) {
-          const subs = (_subFields || []).filter((sub) => subfields.indexOf(sub.tag) >= 0)
-          return subfieldParse(subs)
-          // Otherwise, attempt to return 'content', falling back on subfields' content:
-        } else {
-          return field.content || (_subFields ? subfieldParse(_subFields) : null)
```
nonword, line 184: This legacy logic has grown confusing. The `if(subfields)` check is really a "if querying by specific subfields". Proposing we:
 - isolate (and centralize) the logic around selecting specific subfields either by inclusion or exclusion
 - remove the inline "subfieldParse" function (which has evolved to operate as a "render as varfieldmatch object" function), because it shouldn't be necessary after above
 - generally flatten the structure to clarify how we build the primary value, the subfieldMap, and the parallel prop

lib/sierra-models/base.js
charmingduchess: 👍

lib/sierra-models/base.js
charmingduchess: 👍

lib/sierra-models/base.js
charmingduchess: this was my original idea but i swapped it out for some reason i no longer recall...

lib/sierra-models/base.js
charmingduchess: great this was confusing when i was working on it

charmingduchess: maybe we can call it includedSubfields for symmetry with opts.excludedSubfields?

lib/sierra-models/base.js
charmingduchess: sounds good to me!

lib/sierra-models/base.js
```diff
@@ -147,42 +142,47 @@ class SierraBase {
...
+        // Build parallel:
+        if (field.parallel) {
+          const subfield6 = this._parseSubfield6(field.parallel)
+          // coerce parallel into array for the purpose of recursive call
+          varFieldMatchReturn.parallel = this._prettifyRawVarFields([field.parallel], whichSubfields, opts)[0]
```
charmingduchess, line 178: this function call needs to be updated to the new function name _fieldsWithParallelsAndOrPrimaries

PR #21, opened 2023-03-28
Scc 3422/add m2 indexing

charmingduchess:
This PR incorporates m2 customer code lookups to the indexing process. Largely copied from Paul's work, with an addition for handling a response without a 200 code. @nonword you might want to include something like that in the DHI as well. I'm not sold that this implementation is the move, but there definitely has to be something checking for response.data because 404 responses do not have a data property.
Also tests!

### DIFF
--- a/lib/platform-api/requests.js
+++ b/lib/platform-api/requests.js
@@ -8,19 +8,18 @@ const { chunk } = require('lodash')
  *  Given an array of barcodes, resolves a map relating barcoes to M2 customer
  *  codes (if found).
  */
-const m2CustomerCodesForBarcodes = async (barcodes, offset = 0, map = {}) => {
+const m2CustomerCodesForBarcodes = async (barcodes, offset = 0, map = {}, batchSize = 200) => {
   if (!barcodes || barcodes.length === 0) return {}
-
   // Batch calls to the m2-customer-code-store in groups of 200 barcodes:
-  const batchSize = 200
-  const barcodesBatch = barcodes.slice(offset, offset + batchSize)
 
+  const barcodesBatch = barcodes.slice(offset, offset + batchSize)
   const client = await platformApi.client()
   const response = await client.get(`m2-customer-codes?barcodes=${barcodesBatch.join(',')}`)
 
   // Convert response to a map relating barcode to CC:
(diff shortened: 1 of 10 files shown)


lib/platform-api/requests.js
```diff
@@ -3,9 +3,40 @@ const platformApi = require('./client')
...
+const m2CustomerCodesForBarcodes = async (barcodes, offset = 0, map = {}, batchSize = 200) => {
+  if (!barcodes || barcodes.length === 0) return {}
+  // Batch calls to the m2-customer-code-store in groups of 200 barcodes:
+
+  const barcodesBatch = barcodes.slice(offset, offset + batchSize)
+  const client = await platformApi.client()
+  const response = await client.get(`m2-customer-codes?barcodes=${barcodesBatch.join(',')}`)
+
+  // Convert response to a map relating barcode to CC:
+  const responseAsMap = response.status === 400
```
nonword, line 20: Just curious - did you encounter queries to the m2-customer-codes endpoint that responded with a 400?

test/unit/platform-api-requests.test.js
```diff
@@ -10,32 +10,83 @@ chai.use(require('chai-as-promised'))
...
 describe('platform api methods', () => {
   afterEach(() => {
-    if (platformApi.instance.restore) platformApi.instance.restore()
+    if (platformApi.client.restore) platformApi.client.restore()
     genericGetStub.resetHistory()
     nullGetStub.resetHistory()
   })
+  describe('m2CustomerCodesForBarcodes', () => {
+    it('returns empty object with no barcodes', async () => [
+      await requests.m2CustomerCodesForBarcodes()
```
nonword, line 19: No assertion?

lib/platform-api/requests.js
charmingduchess: i wrote a 400 response for when theres no barcodes or zero customer code matches

test/unit/platform-api-requests.test.js
charmingduchess: whoops. fixed!

nonword (approved):

PR #25, opened 2023-04-07
Add language lookup methods

danamansana:
Adds language lookup methods. Still working on tests.


charmingduchess (approved):
looks ok! pending pauls approval, as well as refactoring the lookup situation

PR #30, opened 2023-04-20
Fixes to parallel note index calculation

nonword:


lib/es-models/bib.js
```diff
@@ -380,54 +385,53 @@ class EsBib extends EsBase {
...
-              index,
-              fieldName: 'note'
-            }
-          })
-      }).flat()
+      // Convert the varfieldmatch objects into parallelDisplayField style
+      // objects:
+      .map(({ varFieldMatchObject, description }, index) => {
+        return {
+          value: parallelValue(varFieldMatchObject),
```
charmingduchess, line 420: 🤘  digging this abstraction

charmingduchess (approved):

PR #31, opened 2023-05-02
SCC 3392: Add EsCheckinCardItem

nonword:
Add EsCheckinCardItem model with supporting utils and tests.

https://jira.nypl.org/browse/SCC-3392


lib/utils/date-parse.js
charmingduchess, line 1: I'm curious why this date parsing looks so different from whats in the discovery store poster. How do we know that the fieldtagv will be compatible with Date()?

nonword: This util is not for parsing dates out of item fieldTag v values, which do vary wildly in format and can not be reliably parsed using `Date`. I'm not sure what shape that will take as code to parse item fieldTag v values has not been incorporated yet afaik. This util is copied directly from [this DSP module](https://github.com/NYPL-discovery/discovery-store-poster/blob/38e5cd559a12e8bcc6ef0f7e2643806a48a00d26/lib/date-parse/date-utils.js#L7)for parsing dates out of checkin card `start_date`, which appear to be ISO 8601 dates with different granularity, making them eminently parsable using this simple util. Should we use a more specific name to make it clear it's for parsing reliable checkin card dates?

lib/es-models/checkin-card-item.js
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    return identifiers
+  }
+
+  /**
+   *  Given a SierraHolding instance, returns an array of EsCheckinCardItem
+   *  instances parsed from it
+   */
+  static fromSierraHolding (sierraHolding) {
```
charmingduchess, line 244: can this be called from the holding and not from within this class? its tricky to wrap my head around.

lib/utils/date-parse.js
charmingduchess: ah ok. yea maybe a comment or a different function name would be useful here.

PR #32, opened 2023-05-04
Wire up the index handler to write ES records

nonword:
Let's wire things up to write to an ES index.

Includes:
 - Slight changes to index handler to process events correctly
 - All ES model classes implement a `.toJson` method that works recursively on children that are ES models
 - Clarified ES client initialization
 - Improved PlatformApiClient initialization (previously it may have been running kms decrypt calls inefficiently)
 - Identifier and log about ES write/connection issues
 - Fix to KMS decryption (wasn't really working before)
 - Small fixes to EsBib methods
 - In general if there are no values to index for a property, the value should be `null`, not `[]` or `['']`
 - Add `./config/` for deployment config
 - Update `sam.local.yml` for writing to a real RCI test index


lib/elastic-search/requests.js
```diff
@@ -4,23 +4,35 @@ const logger = require('../logger')
...
+    resp = await _indexGeneric(process.env.ELASTIC_RESOURCES_INDEX_NAME, records, false)
   } catch (e) {
-    logger.error('writeRecords error: ', e)
+    throw new Error(`Error writing to ES: ${e}`)
+  }
+
+  if (resp.errors) {
+    const errors = resp.items?.map((item) => item.index?.error)
+    let message = 'Unknown error'
+    console.log('handling', resp)
```
charmingduchess, line 19: rm console.log

lib/elastic-search/requests.js
```diff
@@ -4,23 +4,35 @@ const logger = require('../logger')
...
-    logger.debug('Index: ' + (resp.errors ? 'Error' : 'Success'))
-    return resp
+    resp = await _indexGeneric(process.env.ELASTIC_RESOURCES_INDEX_NAME, records, false)
   } catch (e) {
-    logger.error('writeRecords error: ', e)
+    throw new Error(`Error writing to ES: ${e}`)
+  }
+
+  if (resp.errors) {
+    const errors = resp.items?.map((item) => item.index?.error)
```
charmingduchess, line 17: can you put a comment here describing whats going on? looks like elastic search returns errors per item on the response?

lib/es-models/base.js
```diff
@@ -18,6 +19,31 @@ class EsBase {
...
+   *  plainobject by calling .toJson(). If any sub-models implement toJson,
+   *  they will also be rendered as a plainobject.
+   */
+  async toJson () {
+    // Apply to-json to this object:
+    const plainObject = await toJson(this)
+    // Recursively apply to-json to any child entity that implements toJson:
+    await Promise.all(
+      Object.entries(plainObject).map(async ([key, value]) => {
+        // Handle case where ES model method returns a single ES model:
```
charmingduchess, line 34: assuming that the model methods that would return es models are this.items(), this.holdings(), this.checkincards(), and that other es methods wont, can those be called out somewhere in these comments?

test/unit/es-base.test.js
charmingduchess, line 1: maybe there can also be a test where the descendants have toJson as well?

charmingduchess (approved):
Looks good! Just asked for a couple additional comments and a test.

lib/es-models/base.js
nonword: Added [this](https://github.com/NYPL/research-catalog-indexer/pull/32/files#diff-75b3e9ac24d4c013fbdacef132adcdf546bd904a9fa0f3010291ac8462c3f450R27-R30)

PR #34, opened 2023-05-12
Add 856 field parsing

charmingduchess:
- incorporated and edited dsp _extractElectronicResources function
- electronicResources()
- _aeonUrls()
- supplementaryContent()
- numElectronicResources

I based the return values on my notes that i put in the jira ticket for this work, but i'm not 100% on what needs to be in what structure.

### DIFF
--- a/lib/es-models/bib.js
+++ b/lib/es-models/bib.js
@@ -28,9 +28,10 @@ class EsBib extends EsBase {
   }
 
   _aeonUrls () {
-    return this._extractElectronicResourcesFromBibMarc('complete digital record')
-      .map(({ url }) => url)
-      .filter(this._isAeonUrl) || null
+    const urls = this._extractElectronicResourcesFromBibMarc('complete digital record')
+      ?.map(({ url }) => url)
+      .filter(this._isAeonUrl)
+    return urls?.length ? urls : null
   }
 
   carrierType () {
@@ -180,7 +181,7 @@ class EsBib extends EsBase {
 
         return url ? { url, label } : null
       }).filter((r) => r)
-    }
+    } else return null
   }
 
   // _this parameter exists for testing
--- a/test/unit/es-bib.test.js
+++ b/test/unit/es-bib.test.js
@@ -816,12 +816,42 @@ describe('EsBib', function () {
       ])
       expect(electronicResourcesRecord.supplementaryContent()).to.equal(null)
     })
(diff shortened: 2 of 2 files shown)


lib/es-models/bib.js
```diff
@@ -19,6 +19,20 @@ class EsBib extends EsBase {
...
+      'https://specialcollections.nypl.org/aeon/Aeon.dll',
+      'https://nypl-aeon-test.aeon.atlas-sys.com'
+    ]
+    return Boolean(aeonLinks.some((path) => url.startsWith(path)))
+  }
+
+  _aeonUrls () {
+    return this._extractElectronicResourcesFromBibMarc('complete digital record')
+      .map(({ url }) => url)
+      .filter(this._isAeonUrl) || null
```
nonword, line 33: An empty array in Node is truthy, so this will never be `null`

lib/es-models/bib.js
```diff
@@ -533,6 +597,11 @@ class EsBib extends EsBase {
     }
   }
 
+  // Returns array of hashes with { url, label }, containing electronic resources that are not complete digital records
+  supplementaryContent () {
+    return this._extractElectronicResourcesFromBibMarc('supplementary content') || null
```
nonword, line 602: Here also, empty arrays are truth, so this will never fall through to `null`. If you want a one-liner, you need something like
```
  .reduce((ret, el) => (ret || []).concat([el]), null)
```
Or maybe use a `nullIfEmpty` util?

lib/es-models/bib.js
```diff
@@ -123,6 +138,52 @@ class EsBib extends EsBase {
...
+  /**
+ * Given a MiJ bib and a e-resource type, returns extracted resources
+ *
+ * @param {object} bib Instance of BibSierraRecord
+ * @param {string} type Type of resource to extract - either 'supplementary content' or 'complete digital record'
+ *
+ * @return {array} Array of objects representing electronic resources with following properties:
+ * - url: URL of resoruce
+ * - label: Label for resource if found
+ * - path: "Path" to value in marc (for recording provo)
```
nonword, line 156: Should remove this `path` note since we're no longer returning it. Also note possible return of `null` in place of empty array

lib/es-models/bib.js
nonword: I see now. Thanks.

PR #37, opened 2023-05-23
Scc 3389/cli scripts

charmingduchess:
added:
- `reindex-record` behaves just as it did in DHI
- `compare-with-indexed` has a couple add ons:
    - `--activeIndex` which when true will query against the active elastic search index as opposed to the index that this may be wired up to before we deploy. this involved adding another env var too.
    -  `--verbose` which when true will print out a detailed diff with the values of the indexed record and the locally constructed record for comparison
- `currentDocument` to fetch a bib from the elastic search index

changed:
- extracted most of the functionality of `index.js` to `build-es-document.js` for ease of compare-with-index
- add `trimTrailingPunctuation` to the `valueToIndexFromBasicMapping` function
- updated tests to accommodate punctuation changes
- fixed a bug where items were not being added properly to the sierraBib
- fixed typos in strings and made sure values are being returned as arrays from EsBib model functions


nonword (approved):
Looks good

PR #41, opened 2023-06-06
Scc 3390/4/5

danamansana:
This PR rolls up changes for SCC-3390, SCC-3394, and SCC-3395. This includes:
- Replace prefilter method stubs with real implementations. This includes deleting suppressed bibs
- add some null/undefined checks in various methods in `es-models/base`, `es-models/holding`, `es-models/item` 
- add `items` and `holdings` methods in `EsBib`
- add `getSuppressionWithRationale` and related methods to Sierra models
- Add some logic to avoid double-wrapping in `SierraBase`
- Add some tests
- There are still a few commented out tests awaiting implementation
- Made some small changes to existing tests


lib/prefilter.js
charmingduchess, line 1: i feel like the sierra models should be instantiated outside of these prefilter methods. i think it would add clarity to have that happen in the build-es-document file right before the prefilters are called.

lib/platform-api/requests.js
charmingduchess, line 1: is there a way to utilize the prefilterHoldings and Items methods in this file instead of having some of this duplicated code?

lib/sierra-models/bib.js
```diff
@@ -23,6 +25,51 @@ class SierraBib extends SierraBase {
...
+  getSuppressionRationale () {
+    if (this.suppressed) return 'suppressed'
+    if (this.deleted) return 'deleted'
+    if (this.isOtfRecord()) return 'is-otf'
+    return null
+  }
+
+  getSuppressionWithRationale () {
+    const rationale = this.getSuppressionRationale()
+    console.log('rationale: ', rationale)
```
charmingduchess, line 51: rm console.log? or is this supposed to be a logger.info or something?

lib/prefilter.js
```diff
@@ -0,0 +1,52 @@
...
+    .map(bib => new SierraBib(bib))
+    .filter((bib) => {
+      const suppress = bib.isResearch() &&
+        !bib.getSuppressionWithRationale().suppressed
+
+      if (suppress) {
+        suppressBib(bib)
+      }
+
+      return suppress
```
nonword, line 18: Should be `!suppress` ?

lib/prefilter.js
```diff
@@ -0,0 +1,52 @@
...
+
+const prefilterBibs = (bibs) => {
+  return bibs
+    .map(bib => new SierraBib(bib))
+    .filter((bib) => {
+      const suppress = bib.isResearch() &&
+        !bib.getSuppressionWithRationale().suppressed
+
+      if (suppress) {
+        suppressBib(bib)
```
nonword, line 15: Need a way to _not_ issue this DELETE call if we're just building the document for debugging purposes. For example, the `scripts/compare-with-indexed` script invokes `buildEsDocument`, which calls `prefilterBibs`. Perhaps a `options` arg to `buildEsDocument` and `prefilterBibs` that defines a `deleteSuppressedBibs` option, which defaults to `false`. We would set to to `true` when calling `buildEsDocument` from `index.js`..

lib/platform-api/requests.js
```diff
@@ -163,7 +167,11 @@ const _holdingsForBibs = async (bibs) => {
         return null
       }
     }))
-    return holdings.flat()
+
+    return holdings
+      .flat()
+      .map(holding => holding === null ? null : new SierraHolding(holding))
+      .filter(holding => !(holding instanceof SierraHolding && holding.getSuppressionWithRationale().suppressed))
```
nonword, line 174: To avoid these ternaries, we should probably just filter out `null`s since there's no legit reason the `holdings` array should include a `null`

lib/platform-api/requests.js
nonword: As written, it does seem like we could use the prefilter methods, e.g.
```
const _itemsForArrayOfBibs = async (bibs) => {
  try {
    const nyplBibs = bibs.filter((bib) => bib.nyplSource === 'sierra-nypl')
    return Promise.all(nyplBibs.map(async (bib) => {
      const items = await _itemsForOneBib(bib)
      return prefilterItems(items)
    }))
    ...
```
But note that that depends on the current prefilter method implementation wherein we wrap the input in Sierra model classes in order to get access to the `SierraItem.suppressionWithRationale` method. Essentially, we want a clean, side-effect free implementation of our prefilter methods, but the prefilter calculation relies on model methods.

I wonder if some of what feels unstable about this is the uncertainty. Right now we're allowing records to either be wrapped or unwrapped depending on where they are in the pipeline. One way forward may be to simply wrap early and regularly; Don't allow records to hang out unwrapped. Wrap the original event records, run the prefilter, look up bibs (if the event is a holding/item event) and wrap those immediately, do a model prefetch and wrap those immediately.. There would be no need for a separate `buildSierraBibs` method because all logic above that operated exclusively on Sierra model objects..

lib/sierra-models/bib.js
danamansana: Fixed

lib/prefilter.js
danamansana: Fixed

lib/prefilter.js
danamansana: Currently the setup is to use the following check:
```process.env.DISABLE_CIRC_DELETE === 'true'
```
to disable deletion. What if we keep the environment variable approach but change the default behavior to not delete and change the variable name to reflect that? I've made that change but let me know if you don't like it

lib/platform-api/requests.js
danamansana: Fixed

charmingduchess (dismissed):
looks ok to me. the renaming does make the most sense because filtering always happens alongside the sierra model creation. thanks for taking my comments in consideration!

nonword (approved):
Looks good to me!

PR #47, opened 2023-07-18
replace stub

charmingduchess:
We noticed a stubbed function that never got un-stubbed. Mostly copied functions from discovery-api-indexer.


lib/utils/nypl-source-mapper.js
```diff
@@ -4,6 +4,7 @@ class NyplSourceMapper {
   async nyplSourceMapping () {
     if (!this.nyplSourceMap) {
       const resp = await axios.get(`https://raw.githubusercontent.com/NYPL/nypl-core/${process.env.NYPL_CORE_VERSION || 'master'}/mappings/recap-discovery/nypl-source-mapping.json`)
+      console.log(`https://raw.githubusercontent.com/NYPL/nypl-core/${process.env.NYPL_CORE_VERSION || 'master'}/mappings/recap-discovery/nypl-source-mapping.json`)
```
nonword, line 7: Rm console.log

nonword (approved):

charmingduchess:
@nonword just want to double check that a null return for an empty string is correct behavior for dateEndString?

PR #49, opened 2023-07-21
Fixes many things

nonword:
- Add EsBib.parallelTitleDisplay and EsBib.parallelTitleAlt
 - Truncate EsBib.title_sort
 - Fix EsBib.subjectLiteral (and parallelSubjectLiteral) to join subfield values in the unique way we're supposed to join them.
 - Fixes EsBib and EsItem identifier prefixes
 - Fixes EsItem.recapCustomerCode for partner items
 - Fixes EsItem.uri for partner items
 - Fixes items prefetch for partner bibs
 - Adds documentation to EsBase
 - Ensure EsBase.varField returns de-duped VarFieldMatch entries
 - Fix bug linking to parallel fields when primary varField has trailing character data in $6
 - Fix bug in compare-with-indexed
 - Add EsItem.enumerationChronology_sort


lib/utils/subject-formatter.js
charmingduchess, line 1: `truthy` in these function names feels like a red herring. Some clarification about splitting the values into -- delimited and space delimited would be useful!

lib/es-models/bib.js
```diff
@@ -841,6 +854,26 @@ class EsBib extends EsBase {
...
     return [literal.substring(0, 80).toLowerCase()]
   }
+
+  /**
+   *  Get an array of subjectLiteral (or parallelSubjectLiteral) values to index
+   *
+   *  This property has very unusual value joining requirements
+   */
+  _subjectLiteralValueToIndex (primary = true) {
+    const mappings = BibMappings.get('subjectLiteral', this.bib)
```
charmingduchess, line 867: these two lines can be replaced with _valueToIndexFromBasicMapping

lib/sierra-models/base.js
```diff
@@ -160,12 +229,29 @@ class SierraBase {
...
+   * Given an array of raw varField entries from the Sierra record and the
+   * queries sufields and options, returns VarFieldMatch objects
+   *
+   * @return {VarFieldMatch[]}
+   */
   _convertRawFieldToVarFieldMatch (rawFields, includedSubfields, opts) {
     opts = opts || {}
     opts = Object.assign({
-      tagSubfields: false,
       subfieldJoiner: ' ',
```
charmingduchess, line 255: we don't have any subfieldJoiner properties in the mappings anymore. It looks like it was just for the subjectLiteral. I'm happy to see this option workflow be ~~eliminated~~ trimmed down as much as possible. Can we remove the subfieldJoiner logic that remains? I'm also curious if we're using the preFilter option much/at all anymore and if we can get rid of that too?

lib/es-models/bib.js
nonword: But `EsBase._valueToIndexFromBasicMapping` returns an array of strings. I want the `VarFieldMatch` objects (so that I can work with the `subfieldMap` objects directly

charmingduchess: ok!

danamansana (approved):

PR #50, opened 2023-07-28
Scc 3680 addtl

nonword:
## 1. Better deduping of VarFieldMatches

 - We now dedupe VarFieldMatch objects returned from
   SierraBase.varFieldsMulti calls (previously only applied to
   individual SierraBase.varField calls)
 - The de-duping goes to extra effort to identify and remove any
   orphaned parallel that is redundant because another
   VarFieldMatch entry exists with that same value that is also attached
   to a primary value. (Also makes same check for orphaned primaries.

## 2. Fix bug with `excludedSubfields` in mappings

Fixes handling excludedSubfields property of mappings by ensuring that
mapping entries are cast into the proper MarcQuery structure (i.e. put
excludedSubfields into a mappingOptions hash). This bug was causing, for
example, contributorLiteral mappings to add $0 to values because the
excludedSubfields option was being ignored.

### DIFF
--- a/lib/mappings/mappings.js
+++ b/lib/mappings/mappings.js
@@ -26,13 +26,13 @@ const amendMappingsBasedOnNyplSource = (data, nyplSource) => {
  * into SierraBase.varFieldsMulti
  */
 const mappingEntryToMarcQueryObject = (mapping) => {
-  const { marc, subfields } = mapping
-  const mappingOptions = { excludedSubfields: mapping.excludedSubfields }
-  return {
-    marc,
-    subfields,
-    mappingOptions
+  const marcQuery = Object.assign({}, mapping)
+  if (mapping.excludedSubfields) {
+    marcQuery.mappingOptions = { excludedSubfields: marcQuery.excludedSubfields }
+    delete marcQuery.excludedSubfields
   }
+
+  return marcQuery
 }
 
 /**
--- a/lib/mappings/mappings.js
+++ b/lib/mappings/mappings.js
@@ -21,13 +21,41 @@ const amendMappingsBasedOnNyplSource = (data, nyplSource) => {
   }, {})
 }
 
+/**
+ * Given a raw mapping entry, returns a {MarcQuery} object suitable for passing
+ * into SierraBase.varFieldsMulti
+ */
+const mappingEntryToMarcQueryObject = (mapping) => {
+  const { marc, subfields } = mapping
+  const mappingOptions = { excludedSubfields: mapping.excludedSubfields }
+  return {
+    marc,
+    subfields,
+    mappingOptions
+  }
+}
+
+/**
+ *
+ *  @typedef MappingGetter
+ *  @type {function}
+ *  @param {string} propertyName - The property name to look up (e.g. contributorLiteral)
+ *  @param {SierraBase} record - The record we intend to query (which may determine the set of MarcQuery objects returned
+ *  @return {MarcQuery[]}
+ *
+ * Given a record type, returns a function that returns a MappingsGetter function
+ *
+ * @param {string} type - The record type. One of bib, item, or holding
+ * @return {MappingGetter}
+ */
 const makeMappingsGetter = (type) => {
   const mappings = require(`./${type}-mapping.json`)
   return (propertyName, record) => {
     let mappingsForRecord = mappings
     // If nyplSource given, trim mappings to agree with it:
     if (record.nyplSource) mappingsForRecord = amendMappingsBasedOnNyplSource(mappings, record.nyplSource)
-    return mappingsForRecord[propertyName].paths || []
+    return (mappingsForRecord[propertyName].paths || [])
+      .map(mappingEntryToMarcQueryObject)
   }
 }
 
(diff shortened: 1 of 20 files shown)


lib/mappings/mappings.js
```diff
@@ -21,13 +21,41 @@ const amendMappingsBasedOnNyplSource = (data, nyplSource) => {
   }, {})
 }
 
+/**
```
charmingduchess, line 24: is MarcQuery a standard? Or is that just a name you gave this data structure?

lib/sierra-models/base.js
```diff
@@ -156,24 +156,74 @@ class SierraBase {
...
+   *     parallel that also has a primary value
+   *   - Orphaned primary where there exists another entry with an identical
+   *     primary that also has a parallel
+   */
   _uniqueVarFieldMatches (matches) {
     // Create a hash function, which will be used to determine equality of two
     // different maches. We consider two VarFieldMatches as being equal if
     // their primary and parallel values are equal:
     const hasher = (entry) => [entry.value, entry.parallel?.value].join('🎸')
+    let unique = uniqueObjectsByHash(matches, hasher)
```
charmingduchess, line 174: Can you add some documentation to the `uniqueObjectsByHash` function? 
My understanding is that it takes pairs of extracted/transformed values from an array of objects, and the object those are extracted from, and returns only objects for which the extracted/transformed values are unique. Is that accurate?
It also might be useful to describe what is being returned from `uniqueObjectsByHash` using the hasher function provided here.

lib/sierra-models/base.js
```diff
@@ -156,24 +156,74 @@ class SierraBase {
...
+   *       value: v1 parallel
+   *
+   *  .. This function should remove the second entry because it's redundant given the first entry.
+   */
+  _removeRedundantOrphans (matches) {
+    return matches.reduce((a, match) => {
+      // If it's an orphaned parallel
+      const isOrphanedParallel = !!(!match.value && match.parallel?.value)
+      const isOrphanedPrimary = !!(match.value && !match.parallel?.value)
+      const isRedundantOrphan = (isOrphanedParallel || isOrphanedPrimary) && !!matches.find((other) => {
```
charmingduchess, line 208: I like the semantic isOrphaned* variables. Can you create a similar isRedundant using this matches.find logic here?

lib/sierra-models/base.js
```diff
@@ -201,12 +251,18 @@ class SierraBase {
    * @return {VarFieldMatch[]}
    */
   varFieldsMulti (marcObjects) {
-    return marcObjects
+    let matches = marcObjects
       .map(({ marc, subfields, mappingOptions }) => {
-        return this.varField(marc, subfields, mappingOptions)
+        const options = Object.assign({}, mappingOptions, { dedupe: false })
```
charmingduchess, line 256: why are you setting `dedupe: false` if it's set to true in this.varField()?

lib/mappings/mappings.js
nonword: Not a standard. Just giving a name and strict structure to the plainobjects that `varFieldsMulti` expects.

lib/sierra-models/base.js
nonword: But `isRedundant` depends on the VarFieldMatch entity being an orphan; A VarFieldMatch entity has to be an orphan in order for it to be "redundant". I feel that setting up a `isRedundant` var would be misleading because it would sometimes be `true` for things that VarFieldMatches that are not in fact redundant. Unless I restructured to this:

```
const isOrphanedParallel = !!(!match.value && match.parallel?.value)
const isOrphanedPrimary = !!(match.value && !match.parallel?.value)
const isRedundantOrphan = !!matches.find((other) => { 
  // If an identical parallel value is associated with a primary value, remove the orphaned parallel:
  if (isOrphanedParallel) { 
    return other.value && other.parallel?.value === match.parallel.value

  // If an identical primary value is associated with a parallel value, the orphaned primary is redundant:
  } else if (isOrphanedPrimary) { 
    return other.parallel?.value && other.value === match.value
  } 
})
```

It wastes some cycles because we're running that `find` for _all_ VarFieldMatch entities - not just those that have already been identified as either an orphaned parallel or primary. But maybe easier to follow?

charmingduchess: ah ok. makes sense to leave it as is.

lib/sierra-models/base.js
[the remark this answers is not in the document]
nonword: Sure. They're all pretty functional, so don't really need to be class methods. I could create `lib/utils/var-field-match-utils.js`. It would export:
 - uniqueVarFieldMatches
 - parseParallelFieldLink,
 - parseSubField6

.. And would internally use:
 - _removeRedundantOrphans
 
Is that what you have in mind?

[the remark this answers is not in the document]
charmingduchess: yes! exactly. sorry i missed this comment earlier this week.

lib/utils.js
```diff
@@ -67,6 +67,24 @@ const unique = (array) => {
   return Array.from(new Set(array))
 }
 
+/**
+ *  Given an array of plainobjects and a "hasher" function, returns a new
```
charmingduchess, line 71: these look great! thanks for adding.

PR #51, opened 2023-07-28
Additional cleanup

nonword:
- Minor indexing fixes to:
   - EsHolding.notes
   - EsBib.identifierV2 (cancelled/invalid ISBN & ISSN)
 - Fix bug in suppressRecord
 - Better debug log messages in a few places
 - Better display of indexing errors when some but not all failed
 - In compare-with-indexed, don't fail hard if live document isn't found.

### DIFF
--- a/lib/platform-api/requests.js
+++ b/lib/platform-api/requests.js
@@ -55,6 +55,36 @@ const getSchema = async (schemaName) => {
   }
 }
 
+const holdingById = async (id) => {
+  try {
+    const client = await platformApi.client()
+    const resp = await client.get(`holdings/${id}`)
+    if (resp) {
+      return resp
+    } else {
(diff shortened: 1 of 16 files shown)


lib/platform-api/requests.js
```diff
@@ -55,6 +55,36 @@ const getSchema = async (schemaName) => {
...
+      return resp.data
+    } else {
+      logger.warn(`Warning: item not found: ${nyplSource}/${id}`)
+      return null
+    }
+  } catch (e) {
+    logger.error('PlatformApi#itemById error:\n', e)
+  }
+}
+
```
danamansana, line 87: These could probably be refactored since they share a lot of logic but it doesn't seem like a big deal

danamansana (approved):
One small comment but otherwise looks good

PR #52, opened 2023-08-03
Change method for building Aeon URLs

nonword:
No longer require or use bib 856 when building Aeon URLs. Build URL entirely from item and bib data when valid site code is found in item fieldTag x. This is the new way we'd like to build Aeon URLs. The Aeon URLs will differ from the existing 856 sourced URLs in (it is hoped) insignificant ways.

Also centralizes nypl-source-mapper stubbing since a lot of tests will trigger it.

### DIFF
--- a/lib/scsb/requests.js
+++ b/lib/scsb/requests.js
@@ -1,6 +1,6 @@
 const scsb = require('./client')
 const logger = require('../logger')
-const { bNumberWithCheckDigit } = require('../utils')
+const { addSierraCheckDigit } = require('../utils')
 
 // Ensure scsb live querying has not been disabled via ENV:
 const scsbLiveQueryDisabled = process.env.DISABLE_SCSB_LIVE_QUERY === 'true'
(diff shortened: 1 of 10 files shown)


lib/utils.js
```diff
@@ -18,7 +30,7 @@ const bNumberWithCheckDigit = (bnumber) => {
   if (remainder === 11) return `${ogBnumber}0`
   if (remainder === 10) return `${ogBnumber}x`
 
-  return `${ogBnumber}${remainder}`
+  return `${prefix}${ogBnumber}${remainder}`
```
charmingduchess, line 33: Comment on line 16 says 'strip prefix', but you're adding it back here?

charmingduchess:
where were aeon urls being generated before this? manually during cataloguing?

lib/utils.js
nonword: Yeah, we want this function to preserve the prefix (or absence of prefix).

So 
  'b1234' => 'b1234x'
  '1234' => '1234x'

lib/scsb/requests.js
```diff
@@ -15,7 +15,7 @@ const _createRecapCodeMap = async (bib) => {
     const client = await scsb.client()
     const __start = new Date()
     // Bib service returns bibId without check digit, but SCSB requires it, as well as .b prefix
-    const scsbFriendlyBibId = '.b' + bNumberWithCheckDigit(bib.id)
+    const scsbFriendlyBibId = '.b' + addSierraCheckDigit(bib.id)
```
charmingduchess, line 18: is this going to add an extra b after sierraCheckDigit adds the prefix back on?

nonword: No. `addSierraCheckDigit` doesn't add any prefix to that wasn't already there.

charmingduchess: if addSierraCheckDigit returns b1234x, then scsbFriendlyBibId will be .bb1234x. is that the desired behavior?

nonword: `addSierraCheckDigit` will only return "b1234x" if `bib.id` is "b1234". But `bib.id` is "1234", so `addSierraCheckDigit('1234') => '1234x'`

charmingduchess (approved):

PR #55, opened 2023-08-07
Scc 3643/10000 bibs script

charmingduchess:
Create a script to get 10000 (or provided number) bib uris to run with bib-post-runner


lib/utils/nypl-source-mapper.js
```diff
@@ -1,13 +1,8 @@
 const axios = require('axios')
-
+console.log('nypl-source-mapper')
```
nonword, line 2: Rm console.log

nonword (dismissed):

PR #56, opened 2023-08-07
Noref fix suppression of non research

nonword:
Incorporate suppression and is-research checks

 - Add isResearch and supporting methods to Sierra models to ensure we
   don't index Circ records
 - Reorganize build-es-document to ensure we apply is-research and
   suppression checks on bibs regardless of the original event type
 - Update compare-with-indexed script to add support for holding ids and
   inumbers

### DIFF
--- a/lib/sierra-models/bib.js
+++ b/lib/sierra-models/bib.js
@@ -1,7 +1,11 @@
+
 const SierraItem = require('./item')
 const SierraHolding = require('./holding')
 const SierraBase = require('./base')
-const nyplCoreLocations = require('@nypl/nypl-core-objects')('by-sierra-location')
+const {
+  locationHasExclusiveCollectionType,
+  locationHasResearchCenterPrefix
+} = require('../utils/locations')
 
 class SierraBib extends SierraBase {
   constructor (bib) {
@@ -40,9 +44,9 @@ class SierraBib extends SierraBase {
   }
 
   getSuppressionRationale () {
-    if (this.suppressed) return 'suppressed'
-    if (this.deleted) return 'deleted'
-    if (this.isOtfRecord()) return 'is-otf'
+    if (this.suppressed) return 'Suppressed'
+    if (this.deleted) return 'Deleted'
+    if (this.isOtfRecord()) return 'Is OTF'
     return null
   }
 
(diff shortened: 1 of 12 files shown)


lib/sierra-models/bib.js
```diff
@@ -51,23 +55,51 @@ class SierraBib extends SierraBase {
...
+      return { isResearch: true, rationale: 'Locations include none/os' }
+    } else {
+      // Determine locations that are exclusively 'Branch'
+      const branchLocations = locationCodes
+        .filter((code) => locationHasExclusiveCollectionType(code, 'Branch'))
+      // If bib has any locations with just "Branch" collection type, bib is not research:
+      if (branchLocations.length > 0) {
+        return {
+          isResearch: false,
+          rationale: `Has branch locations: ${branchLocations.join(', ')}`
```
charmingduchess, line 79: are there any bibs with some research and some circ items?

nonword: No. Or if there are, we don't support then.

test/unit/utils-locations.test.js
```diff
@@ -0,0 +1,34 @@
...
+describe('utils/locations', () => {
+  describe('locationHasExclusiveCollectionType', () => {
+    it('returns true for location with matching collectionType', () => {
+      expect(locationHasExclusiveCollectionType('mal92', 'Research')).to.equal(true)
+      expect(locationHasExclusiveCollectionType('aga01', 'Branch')).to.equal(true)
+    })
+
+    it('returns false for location with non-matching collectionType', () => {
+      expect(locationHasExclusiveCollectionType('mal92', 'Branch')).to.equal(false)
+      expect(locationHasExclusiveCollectionType('aga01', 'Resaerch')).to.equal(false)
```
charmingduchess, line 17: though it doesn't matter for the test to pass, can you fix this typo for clarity of what is being tested here?

lib/sierra-models/item.js
```diff
@@ -31,6 +39,55 @@ class SierraItem extends SierraBase {
     return type
   }
 
+  getIsResearchWithRationale () {
```
charmingduchess, line 42: I think it would be useful to summarize the checks that are being made here.

nonword: I refactored a bit and added a bunch of documentation.

charmingduchess: 🤩 looks lovely!

charmingduchess (approved):
Thanks for refactoring - much easier to scan now!

PR #58, opened 2023-08-11
Noref updated suppression rules

nonword:
Fix many things around suppression and is-research

 - Light cleanup of build-es-document.js
 - Updated Bib suppression check. A bib is Research if:
   - is a partner record OR
   - has a 910|a of RL/RLOTF OR
   - has locations tagged Research in NYPL-Core OR
   - has locations whose prefixes look like Research location prefixes
 - Improve reporting in compare-with-indexed around whether/why
   something is filtered out due to suppression or Research check
 - Added additional Aeon site codes from Antonio
 - Simplified (made more direct) the method for linking SierraBibs to
   their child -Holdings and -Items
 - Updated QA SCSB config
 - Remove buildSierraBibs from build-es-document because it wasn't really serving its name any more given that all records are wrapped immediately. The one functional purpose it served any more was to add back-references to the parent SierraBib of SierraItems and SierraHoldings. This purpose has been re-sited in the modelPrefetch function, since that's where we're currently establishing that hierarchy.
 -  Ensure bibId lookups de-dupe bib ids so that an event that contains multiple items/holdings that belong to the same bibs only process common bibs once.


lib/build-es-document.js
```diff
@@ -46,17 +43,37 @@ const buildEsDocument = async ({ type, records }) => {
...
+const transformIntoBibRecords = async (type, records) => {
+  // If incoming event is Item/Holding, convert it into a Bib event:
+  if (type === 'Item') {
+    records = await filteredSierraItemsForItems(records)
+    records = await platformApi.bibsForHoldingsOrItems(type, records)
+  } else if (type === 'Holding') {
+    records = await filteredSierraHoldingsForHoldings(records)
+    records = await platformApi.bibsForHoldingsOrItems(type, records)
+  }
+  return records
```
charmingduchess, line 76: is there a chance that records may contain duplicate bibs we'd want to filter out?

lib/build-es-document.js
```diff
@@ -46,17 +43,37 @@ const buildEsDocument = async ({ type, records }) => {
 
 const buildSierraBibs = (records) => {
```
charmingduchess, line 44: is this being called anywhere anymore?

nonword: Good catch! Technically we don't need the reverse references any more, so it didn't break anything, but you're right I broke that wiring. Will fix.

lib/build-es-document.js
nonword: Good idea. I'll make sure that's happening and add some test coverage

lib/build-es-document.js
charmingduchess: why don't we need the reverse references?

lib/build-es-document.js
nonword: This is really controlled by `_bibIdentifiersForHoldings` and `_bibIdentifiersForItems`. I updated the former to ensure it de-dupes bib identifiers. The latter was already de-duping, but I cleaned it up a little. Added specific tests for both.

lib/build-es-document.js
nonword: Actually I was thinking of ES model back references, which are newly no longer needed because EsItem.aeonUrl no longer depends on anything in the parent EsBib. In the case of Sierra models, I'm not aware of any historical need, although i still think it's a sensible thing to have ready.

charmingduchess: Given that we don't actively need this behavior now, maybe we don't include it?

nonword: If it wasn't already built and tested, we might opt not to build it, but as it is built and tested I would vote to include it. I also just think reverse model traversal is a sensible feature of a data model class like this. Maybe not essential enough to build it without an immediate need, but enough so to include it if it's already there.

charmingduchess (approved):

PR #60, opened 2023-09-14
Fixes

nonword:
Includes several fixes:
 - Updates to Aeon URL generation
 - Ensure EsBib.carrierType and EsBib.mediaType can have multiple entities when multiple values are parsed
 - Fix to shelfMark_sort fallback
 - Fall back on bib shelfMark when none found in item
 - Implement custom subfieldJoiner for genreForm (need to be joined with an en-dash)
 - Fix to Sierra checksum generation for some identifiers
 - Added production config
 - Trim trailing punctuation earlier in marc query to ensure varfield de-duping is applied to the final values


lib/es-models/bib.js
```diff
@@ -807,6 +809,8 @@ class EsBib extends EsBase {
   }
 
   _isAeonUrl (url) {
+    if (typeof url !== 'string') return false
```
charmingduchess, line 812: is this a case that you've seen or just extra caution?

lib/mappings/mappings.js
```diff
@@ -27,10 +27,15 @@ const amendMappingsBasedOnNyplSource = (data, nyplSource) => {
  */
 const mappingEntryToMarcQueryObject = (mapping) => {
   const marcQuery = Object.assign({}, mapping)
-  if (mapping.excludedSubfields) {
-    marcQuery.mappingOptions = { excludedSubfields: marcQuery.excludedSubfields }
-    delete marcQuery.excludedSubfields
-  }
+  // If either of these properties are set in the mappings json, copy them to the mappingOptions:
+  ; ['excludedSubfields', 'subfieldJoiner']
```
charmingduchess, line 31: for some reason i really love a leading semicolon... 👍🏻

charmingduchess: but on a more serious note - the following logic is a bit hard to parse because the variables have such similar names. can you summarize briefly in a comment?

lib/sierra-models/base.js
```diff
@@ -349,6 +351,11 @@ class SierraBase {
           varFieldMatchReturn.value = field.content
         }
 
+        // Trim trailing punctuation:
+        if (varFieldMatchReturn.value) {
+          varFieldMatchReturn.value = trimTrailingPunctuation(varFieldMatchReturn.value)
```
charmingduchess, line 356: this seems like a good relocation. what inspired it?

lib/utils.js
charmingduchess, line 1: this code is in a couple places, i think. was this a bug we need to address elsewhere or just clarity?

charmingduchess (approved):
Just one request for comment and a couple questions for my own edification!

lib/es-models/bib.js
nonword: Should now be protected upstream, but good to check types before calling 'startsWith'

lib/sierra-models/base.js
nonword: Saw a duplicate note come through because our previous de-duping was being applied to the original varfield values and only later trimming trailing punctuation. This meant that trailing punctuation caused two different varfield values to appear as unique, but then later appear identical because of the late application of trimTrailingPunctuation. By trimming before de-duping, we fix this.

lib/utils.js
nonword: If this refers to `addSierraCheckDigit`, yes it addresses a bug. Previously we were dropping the prefix when remainder was 11 or 10

lib/mappings/mappings.js
nonword: Added some inline comments to try to clarify

PR #61, opened 2023-09-19
Noref fix deletions

nonword:
Fix bug where indexer failed to delete anything due to an apparent bug in latest `elasticsearch` library (probably a support gap for our very old ES version).

Also fixes tiny checkin card format parsing bug.

### DIFF
--- a/lib/elastic-search/requests.js
+++ b/lib/elastic-search/requests.js
@@ -107,10 +107,10 @@ const currentDocument = async (bibId, index) => {
   const record = await client.search({
     index,
     body: {
-      query: { term: { uris: bibId } }
+      query: { term: { uri: bibId } }
     }
   })
-  if (!record.hits.hits[0]) throw new Error(`Could not find ${bibId} in ElasticSearch`)
+  if (!record.hits.hits[0]) throw new Error(`Could not find ${bibId} in ${index}`)
   return record.hits.hits[0]._source
 }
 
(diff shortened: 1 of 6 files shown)


lib/utils/suppressBibs.js
```diff
@@ -6,16 +6,15 @@ const suppressBib = async (bib) => {
...
+  logger.debug(`Deleting: ${bibUri}`)
 
-  return elasticClient.delete({
+  const opts = {
     index: process.env.ELASTIC_RESOURCES_INDEX_NAME,
     type: 'resource',
     id: bibUri
-  })
+  }
+  return elasticClient.delete(opts)
```
charmingduchess, line 17: is this not an async call? do we want to await it?

lib/elastic-search/requests.js
```diff
@@ -107,10 +107,10 @@ const currentDocument = async (bibId, index) => {
   const record = await client.search({
     index,
     body: {
-      query: { term: { uris: bibId } }
+      query: { term: { uri: bibId } }
```
charmingduchess, line 110: was this the bug?

nonword: We no longer index `uris` (containing both bnum and all item nums), bug we've always indexed `uri` (just the bnum), so anywhere we're querying `uris` needs to change to `uri`

lib/utils/suppressBibs.js
nonword: It is an async call, but we don't need to do anything with the resolved value in this function, so it's fine to return the unresolved Promise

charmingduchess (approved):

PR #62, opened 2023-09-20
Add ItemVolume to aeon URL

nonword:
By request, adding a new ItemVolume= param to Aeon URLS to hold volume information. (Removing it from CallNumber)

### DIFF
--- a/lib/utils/aeon-urls.js
+++ b/lib/utils/aeon-urls.js
@@ -1,20 +1,9 @@
 const { firstValue, addSierraCheckDigit, trimTrailingPunctuation } = require('../utils')
 const logger = require('../logger')
+const { truncate } = require('../utils')
 const { lookup } = require('./lookup')
 
 const locationLabelMap = lookup('aeon-site-code-to-label')
-
-/**
- *  Simple util for truncating a string to the specified length and adding a
- *  ellipse suffix indicating that a truncation has occurred.
- */
-const truncate = (s, length) => {
-  if (!s || s.length <= length) return s
-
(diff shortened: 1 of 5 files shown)


lib/utils/aeon-urls.js
```diff
@@ -37,7 +26,8 @@ const aeonUrlForItem = async (esItem) => {
 
   const baseUrl = aeonBaseUrl + '/aeon/Aeon.dll?'
 
-  const sierraBib = esItem.item.bibs() && esItem.item.bibs().length ? esItem.item.bibs()[0] : null
+  const sierraItem = esItem.item
```
charmingduchess, line 29: while we're in the business of clearing up this ternary logic, maybe `sierraItem.bibs()?.length` will assist?

lib/utils/aeon-urls.js
```diff
@@ -56,11 +46,12 @@ const aeonUrlForItem = async (esItem) => {
   const locationLabel = locationLabelMap[siteCode]
   const holdingLocationId = firstValue(esItem.holdingLocation())?.id?.split(':').pop()
   const isReCAP = /^rc/.test(holdingLocationId)
+  const itemVolume = sierraItem.fieldTag('v')[0]?.value
```
charmingduchess, line 49: why not use the esItem.shelfMark(true) here?

lib/utils/aeon-urls.js
nonword: Good call

nonword: Fixed

charmingduchess (approved):

PR #63, opened 2023-09-21
Final QA fixes

nonword:
Fixes a number of things that came up in testing and deployment:
 - Order items by enumerationChronology_sort desc (not essential, but a nicety for debugging since that's how they're displayed)
 - Update QA config to use live index
 - Update config to use INFO logging because DEBUG is super chatty and should only be used when debugging
 - Fix a couple bad identifier type namespaces


config/qa.env
charmingduchess, line 1: 🥳

charmingduchess (approved):

PR #67, opened 2023-10-24
Fix bug identifying newer rc locations as Research

nonword:
Updates the location prefix check to ensure that we assume items in any "rc" prefixed location are ReCAP even if we don't yet have them in NYPL-Core.

Also adds several tests to improve coverage of the `getIsResearchWithRationale` call.


charmingduchess:
Can you write a test using one of the items that was erroneously suppressed? or some object with the same relevant fields?

PR #72, opened 2023-11-04
Scc 3794

nonword:
Write a brief record to `IndexDocumentProcessed` kinesis stream after writing to ES - for easy monitoring

https://jira.nypl.org/browse/SCC-3794


index.js
```diff
@@ -19,8 +20,12 @@ const handler = async (event, context, callback) => {
 
     let message = ''
     if (recordsToIndex.length) {
+      // Write records to ES:
       await elastic.writeRecords(recordsToIndex)
 
+      // Write to IndexDocumentProcessed Kinesis stream:
+      await notifyDocumentProcessed(recordsToIndex)
```
danamansana, line 27: Not a big deal but since it doesn't look like we are doing anything with this value does it need to be `await`ed?

danamansana (approved):
Looks good just one small comment

index.js
nonword: Yes. `notifyDocumentProcessed` does async work. We don't want `handler` to return until that work is done. When we callback from `handler`, we're telling the Lambda environment that work is complete. If there is lingering async work happening that no one is awaiting, we can't be sure it will finish before the environment is destroyed.

PR #80, opened 2024-03-06
Noref fix mediatype bug

nonword:
Fix production bug encountered in partner record. Indexer assumed a 337 would have primary content, whereas some only have parallel content and should be ignored.


charmingduchess (approved):
how strange. whats the item that triggered the bug?

nonword:
> how strange. whats the item that triggered the bug?

PUL bib 9953516043506421 has an 880 for 337, but no primary 337

PR #82, opened 2024-03-29
Add CSV support to bulk-index script

nonword:
Updated bulk-index script to support --csv and related args. Script can process a CSV of almost any length containing a mix of columns. The identifier is expected to be a prefixed identifier (e.g. "b1234", "pb4567") and may represent multiple different types and nyplSource's. Script processes records in 50-100K/h and reports rolling estimated ETA.

Also update winston logger package and wrapper for simplicity and easier local logging.


scripts/bulk-index.js
```diff
@@ -231,6 +265,10 @@ modelPrefetcher.modelPrefetch = async (bibs) => {
     bib._holdings = filteredSierraHoldingsForHoldings(holdingsByBibId[bib.id] || [])
     // Apply suppression/is-research filtering to items:
     bib._items = filteredSierraItemsForItems(itemsByBibId[bib.id]) || []
+    // Ensure items have a reference to their bib:
+    bib._items.forEach((item) => {
+      item._bibs = [bib]
```
charmingduchess, line 270: Are we not worried about bound with items here?

scripts/bulk-index.js
charmingduchess, line 1: Can you write unit tests for batchBySize? and maybe also updateByCsv (if its not too annoying to mock)?

scripts/bulk-index.js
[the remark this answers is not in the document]
nonword: Not yet, no. We do intentionally link to the parent via the plural `_bibs` property because eventually it would be nice to work out how to better support bound-with. But from the index and discovery-api perspective, bound-with items are just items that are duplicated across multiple bibs; Bibs are the top-level model. In any case, this is really just a performance optimized reimplementation of [the non-bulk model prefetch](AQECAHh7ea2tyZ6phZgT4B9BDKwguhlFtRC6hgt+7HbmeFsrsgAAAHIwcAYJKoZIhvcNAQcGoGMwYQIBADBcBgkqhkiG9w0BBwEwHgYJYIZIAWUDBAEuMBEEDMK4A48qmN1m/MwZWgIBEIAvuuag76L0g0c2C+Jm4wiX3bLNpix2m4IPuEQLTgQ/eitS1QDpDFRaU9rGSccdvGQ=), which does a similar simplification on bound-with items and we need to match that (or else bulk-indexed records will differ from pipeline-indexed records)

scripts/bulk-index.js
nonword: Good call. Added test coverage over those and a few other bulk-index concerns.

charmingduchess (approved):

PR #84, opened 2024-04-13
ES v8 support: Update ES module, fixes to tests

nonword:
Readying RCI to write to the new ES v8 domain:
 - Update to latest ES client module
 - Support extracting username & pass from encrypted connection config
 - Adds `config/qa-new-domain.env` for writing new v8 domain
 - Update existing qa and prod connection config to use `https://` proto for consistency
 - Small app/test updates for new ES client api


danamansana (commented):
Looks good but I think there are some conflicts to resolve. How is `qa-new-domain.env` going to be used in practice?

nonword:
Fixed the upstream conflicts.

The `qa-new-domain.env` file is just there to document the new ES domain config - without actually activating it. It's there so that the info doesn't live exclusively on my computer. This branch adds "support for" the new ES v8 domain without actually activating it in QA. We need to finish validating the v8 domain in QA before switching over to the new domain officially (which we'll want to do in concert with a change to the discovery-api and possibly only after we have a production v8 ES in place).

danamansana (approved):

PR #85, opened 2024-04-16
SCC 4059: Add EsBib.addedAuthorTitle

nonword:
Adds new EsBib property for capturing the title of a work associated with a named creator.

Also:
 - Ports the `mapping-check` over from discovery-api-indexer to assist identifying field mapping differences between this app and any remote index
 - Adds a couple properties found to be missing from the index-schema doc (via testing above)

To avoid doing this twice, this PR builds on [this PR](https://github.com/NYPL/research-catalog-indexer/pull/84), which adds support for v8 and updates the elasticsearch client.


lib/mappings/bib-mapping.json
```diff
@@ -6,10 +6,13 @@
       }
     ]
   },
-  "aeonURL": {
```
nonword, line 9: We're not removing the `aeonURL` property, but we don't rely on this mapping file to build it. Just cleaning up.

lib/elastic-search/index-schema.js
```diff
@@ -183,6 +187,7 @@ exports.schema = () => ({
       },
       idBarcode: propertyTemplates.exactString,
       location: propertyTemplates.entity,
+      m2CustomerCode: propertyTemplates.exactString,
```
charmingduchess, line 190: were these fields just not being indexed till now?

test/unit/es-index-schema.test.js
```diff
@@ -5,6 +5,7 @@ describe('elastic search schema', () => {
   it('returns the schema', () => {
     const properties = Object.keys(schema())
     expect(properties).to.include('carrierType')
-    expect(properties.length).to.equal(88)
+    // At last count, we have 89 bib-level field mappings:
```
charmingduchess, line 8: Is this comment supposed to not match the number being tested?

lib/elastic-search/index-schema.js
nonword: They were being indexed; They were not mentioned in the `index-schema` file, which wasn't really being used yet.

test/unit/es-index-schema.test.js
nonword: The test as written was brittle because it breaks every time we add a prop. The test now asserts that the number of bib properties exceeds 88, which as of this change it does.

charmingduchess (approved):
Looks good. Did the mapping check have any tests in the discovery api indexer? Might be a good idea to have even one integration test (on the `run` function) for future devs who might want to change the code.

nonword:
I added test coverage over the `mapping-check` script and merged in latest from `main`. The diff is now super chatty about things that are not actually diffs. Hoping when [this](https://github.com/NYPL/research-catalog-indexer/pull/84) is approved and merged, it's easier to re-review.

lib/elastic-search/index-schema.js
```diff
@@ -131,7 +132,7 @@ exports.schema = () => ({
           label: propertyTemplates.exactString
         }
       },
-      note: propertyTemplates.exactString,
+      notes: propertyTemplates.exactString,
```
nonword, line 135: This fixes a typo

charmingduchess (approved):

PR #90, opened 2024-07-18
A few added fields for ES tuning

nonword:
Adds a couple new ES fields to improve author matching.

Also fixes a couple minor things with bulk-index and compare-with-indexed scripts.

~This depends on Node20 updates, so targets that branch currently, but will be merged into `main` when which are now in `main` since [PR 89](https://github.com/NYPL/research-catalog-indexer/pull/89) is merged.~

### DIFF
--- a/lib/utils/author-names.js
+++ b/lib/utils/author-names.js
@@ -1,0 +1,50 @@
+// Match the following patterns:
+//  - Lastname, (Firstname|Iniital.)
+//  - Lastname, (Firstname|Initial.) (Middlename|Initial.)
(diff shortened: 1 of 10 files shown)


lib/utils/author-names.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+  }
+
+  return permutations
+}
+
+const withoutDates = (name) => {
+  if (typeof name !== 'string') return name
+
+  const tokens = name.split(', ')
```
charmingduchess, line 41: I'm struggling to find an example, but I imagine there could be a case of a mononymous author with dates?

charmingduchess (approved):
Looks good, just pointed out an edge case that might need addressing.

lib/utils/author-names.js
nonword: Cher!
```
"creatorLiteral": "Cher, 1946-"
```

Good catch. Fixed

charmingduchess: That's so funny. She was the only person I could think of without googling a list of authors and I never thought to try her.

charmingduchess (approved):

PR #92, opened 2024-09-19
Scc 4268

nonword:
Add support for using --hasSubfield when using --hasMarc. Allows one to reindex records that have a specific marc and subfield combination.

Updates to tests. 

Fix to qa config.

https://qa-platform.nypl.org/docs/check-in-cards


scripts/bulk-index.js
```diff
@@ -400,22 +406,40 @@ const buildSqlQuery = (options) => {
...
+    }
+
+    sqlFromAndWhere = selects.join(',\n') +
+      '\nWHERE ' + wheres.join('\nAND ')
   } else {
     throw new Error('Insufficient options to buildSqlQuery')
   }
 
-  const query = `SELECT * FROM ${sqlFromAndWhere}` +
+  const innerSelect = `SELECT DISTINCT id, nypl_source FROM ${sqlFromAndWhere}` +
```
charmingduchess, line 436: can you add a comment here describing why this is necessary?

charmingduchess (approved):
Looks good!

scripts/bulk-index.js
nonword: Added note about how and why.

charmingduchess (approved):

PR #93, opened 2024-09-24
Ensure contributor* fields don't duplicate creator*

nonword:
Updates contributorLiteralNormalized and contributorLiteralWithoutDates fields to remove entries that are found to duplicate creator* fields.

This is to ensure we don’t ever have a bib with, for example, creatorLiteralNormalized "Charles Dickens" and contributorLiteralNormalized "Charles Dickens" due to the two different Charles's being named in different roles, which would unfairly boost this record when matching on both of those Normalized fields.

Also adds removal of trailing period characters in creator and contributor fields so that we don't ever index e.g. "Dickens, Charles.".

https://newyorkpubliclibrary.atlassian.net/browse/SCC-4278

### DIFF
--- a/lib/es-models/bib.js
+++ b/lib/es-models/bib.js
@@ -121,22 +121,24 @@ class EsBib extends EsBase {
   * All distinct contributor names after "normalization" - `Firstname Lastname`
   */
   contributorLiteralNormalized () {
-    const names = this.contributorLiteral()
+    let names = this.contributorLiteral()
     if (!names) return null
 
-    return names.map(normalizeAuthorName)
+    names = names.map(normalizeAuthorName)
       .flat()
       // Remove any names that duplicate normalized creator names:
       .filter((contrib) => !nameIsRedundant(contrib, this.creatorLiteralNormalized()))
+    return names.length === 0 ? null : names
   }
 
   contributorLiteralWithoutDates () {
-    const names = this.contributorLiteral()
+    let names = this.contributorLiteral()
     if (!names) return null
 
-    return names.map(withoutDates)
+    names = names.map(withoutDates)
       // Remove any names that duplicate creator names w/out dates:
       .filter((contrib) => !nameIsRedundant(contrib, this.creatorLiteralWithoutDates()))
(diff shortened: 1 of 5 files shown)


lib/es-models/bib.js
```diff
@@ -101,32 +106,37 @@ class EsBib extends EsBase {
...
-    const creators = this.creatorLiteral()
-    if (!creators) return contributorLiterals
-    contributorLiterals = contributorLiterals.filter((contrib) => {
+    contributorLiterals = contributorLiterals
+      .map(trimTrailingPeriod)
       // Contributor is redundant if there exists a creator of which the
       // contributor is a prefix/match.
       // E.g. contributor "Dickens, Charles" is redundant if there's a creator
       // with "Dickens, Charles" or "Dickens, Charles, 1812-1870"
-      return !creators.find((creator) => creator.startsWith(contrib))
```
nonword, line 112: Lines 104-112 are now generalized in the `nameIsRedundant` util

lib/es-models/bib.js
```diff
@@ -101,32 +106,39 @@ class EsBib extends EsBase {
     let contributorLiterals = this._valueToIndexFromBasicMapping('contributorLiteral')
     if (!contributorLiterals) return null
 
-    // Filter out any contributors that also appear as creator:
-    const creators = this.creatorLiteral()
-    if (!creators) return contributorLiterals
-    contributorLiterals = contributorLiterals.filter((contrib) => {
+    contributorLiterals = contributorLiterals
+      .map(trimTrailingPeriod)
       // Contributor is redundant if there exists a creator of which the
```
charmingduchess, line 111: I think this comment can be removed

charmingduchess (changes requested):
looks good! just a couple updates requested

charmingduchess (approved):

lib/es-models/bib.js
nonword: I wanted some explanation about what is happening in that filter, so I simplified the comment.

PR #95, opened 2024-10-15
Activate v8 index, plus extras

nonword:
Activates RCI on latest QA v8 index:
 - Update qa config and local sam file
 - Update ES client wrapper to parse and use updated authorization method

Other small fixes:
 - Update to-json method to ensure keys are alpha-sorted
 - Fix bug where partner bibs weren't getting buildingLocationIds

 https://newyorkpubliclibrary.atlassian.net/browse/SCC-4310


charmingduchess (approved):
Looks good. where do the new ES username and passwords come from?

nonword:
> Looks good. where do the new ES username and passwords come from?

We support authing by username & password as part of this (if found in the connection URI), but we're now authing by api-key. API keys were created in Kibana by Devops and encrypted as `ENCRYPTED_ELASTICSEARCH_API_KEY`

PR #104, opened 2024-12-17
SCC-4347 - Add scripts/update-property-by-csv script

nonword:
Add update-property-by-csv script for updating a single property in many ES documents based on a CSV of any length. For example, this can be used to update the `recordTypeId` for every bib in the index if you have a CSV containing columns id, nyplSource, recordTypeId.

https://newyorkpubliclibrary.atlassian.net/browse/SCC-4347


scripts/update-property-by-csv.js
```diff
@@ -0,0 +1,318 @@
...
+    }
+  })
+}
+
+/**
+* Given a ES _bulk response, analyzes the response for errors. Returns an object that defines:
+*  - errored {object[]} - Array of objects representing non-404 errors, one per document
+*  - missing {object[]} - Array of objects representing 404 errors, one per document
+**/
+const parseErroredDocuments = (bulkResponse, operations) => {
```
charmingduchess, line 142: what is operations here?

nonword: Added documentation. It's the array of objects we send to the ES _bulk api that alternately identify the document to update and the property to update. E.g.:

```
   [
        { update: { _index: 'index-name', _id: 'b1' } },
        { doc: { myProp: 'new prop value 1' } },
        { update: { _index: 'index-name', _id: 'b2' } },
        { doc: { myProp: 'new prop value 2' } },
        ...
   ]
```

Sending a stream of those in one call to the ES _bulk api is very fast compared with issuing POSTs to each document id

charmingduchess (approved):

PR #105, opened 2025-01-17
[SCC-4455] Support multiple 910|a fields

yossariano:
See https://newyorkpubliclibrary.atlassian.net/browse/SCC-4455 - this change adds support for checking multiple `910|a` values for whether a bib/item is meant for research, instead of just the first value in the field.

#### Testing
* Updated tests to include a fixture with both `BL` and `RL` in the `910|a` field.

### DIFF
--- a/lib/sierra-models/bib.js
+++ b/lib/sierra-models/bib.js
@@ -35,9 +35,9 @@ class SierraBib extends SierraBase {
     // https://github.com/NYPL/nypl-core/blob/master/vocabularies/csv/locations.csv#L2265
     const hasOtfLocation = this.locations && this.locations[0] && this.locations[0].code === 'os'
 
-    // Having 910 $a === 'RLOTF' is also a solid indicator introduced in OTF
+    // Having 910 $a contain 'RLOTF' is also a solid indicator introduced in OTF
     // record creation Oct 2021 https://github.com/NYPL/recap-hold-request-consumer/pull/33
-    const hasOtf910 = this._varField910() === 'RLOTF'
+    const hasOtf910 = this._varField910().includes('RLOTF')
 
     return hasOtfLocation || hasOtf910
   }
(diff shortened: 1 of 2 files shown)


lib/sierra-models/bib.js
```diff
@@ -103,14 +103,14 @@ class SierraBib extends SierraBase {
...
 
   /**
-   *  Get the value in 910 $a if one exists.
+   *  Get the values in 910 $a.
    *
-   *  This value typically has RL for Research bibs, BL for Branch bibs, or
+   *  These values typically are RL for Research bibs, BL for Branch bibs, or
    *  RLOTF for OTF (Research) bibs
    */
   _varField910 () {
```
charmingduchess, line 111: this is not technically in scope, but could you update this method name to something like `_varField910a`?

lib/sierra-models/bib.js
```diff
@@ -70,10 +70,10 @@ class SierraBib extends SierraBase {
     }
 
     // Check 910 $a for RL or BL:
-    const researchBranchFlag = this._varField910()
-    if (researchBranchFlag) {
-      const isResearch = ['RL', 'RLOTF'].includes(researchBranchFlag)
-      return { isResearch, rationale: `910 $a is ${researchBranchFlag}` }
+    const researchBranchFlags = this._varField910()
+    if (researchBranchFlags.length !== 0) {
+      const isResearch = researchBranchFlags.includes('RL') || researchBranchFlags.includes('RLOTF')
```
charmingduchess, line 75: is this just a stylistic preference?

charmingduchess (changes requested):

charmingduchess:
I'm also not seeing any prior unit tests for this method. I think it would be a good idea to add some edge case coverage for `isOtfRecord` or maybe updating [these suppression tests](https://github.com/NYPL/research-catalog-indexer/blob/SCC-4455-multiple-910a/test/unit/sierra-bib.test.js).

lib/sierra-models/bib.js
yossariano: Yeah, I'm open to any style suggestions!

lib/sierra-models/bib.js
yossariano: Absolutely, will update.

yossariano:
> I'm also not seeing any prior unit tests for this method. I think it would be a good idea to add some edge case coverage for `isOtfRecord` or maybe updating [these suppression tests](https://github.com/NYPL/research-catalog-indexer/blob/SCC-4455-multiple-910a/test/unit/sierra-bib.test.js).

It is implicitly tested with this test (it failed with multiple `910|a` fields previously):

```
  prefilter
    filteredSierraBibsForBibs
      1) should keep Research bibs
```

but I'm adding more explicit testing now!

nonword (approved):

PR #107, opened 2025-01-29
Add identify-ids-by-es-query script

nonword:
A script for building a CSV of bib ids for an arbitrary ES query. Useful for building a CSV of records for feeding into the RCI bulk-index script.

### DIFF
--- a/scripts/identify-ids-by-query.js
+++ b/scripts/identify-ids-by-query.js
@@ -66,10 +66,12 @@ const parseArguments = () => {
         type: 'string',
         default: '0'
       },
-      outfile: { type: 'string' },
-      query: {
-        type: 'string'
+      outfile: {
+        type: 'string',
+        default: './out.csv'
       },
+      limit: { type: 'string' },
+      query: { type: 'string' },
       size: {
         type: 'string',
         default: '100'
@@ -81,7 +83,7 @@ const parseArguments = () => {
     }
   })
 
-  castArgsToInts(argv, ['batchSize', 'limit', 'offset'])
+  castArgsToInts(argv, ['limit'])
 
   return argv.values
 }
--- a/scripts/identify-ids-by-query.js
+++ b/scripts/identify-ids-by-query.js
@@ -1,0 +1,189 @@
+/**
+ *
+ * Given a ES query, identfies matching records and writes them to a local csv
+ *
+ * Options:
+ *  --query QUERY - Provide ES query as a quoted JSON blob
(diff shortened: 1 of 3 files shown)


charmingduchess (approved):
This looks good! 
I had initially come to the [scroll api](https://www.elastic.co/guide/en/elasticsearch/reference/current/scroll-api.html) when I was starting my own version of this script, but was diverted by this alert in the docs:

> We no longer recommend using the scroll API for deep pagination. If you need to preserve the index state while paging through more than 10,000 hits, use the [search_after](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#search-after) parameter with a point in time (PIT).

I don't imagine that this is a problem for this script, but something to keep in mind for updates in the future. I have a feeling that the search_after might also help with the bug where duplicated results turn up. I can't link to this line, but here's what caught my eye:

> Repeat this process by updating the search_after array every time you retrieve a new page of results. If a [refresh](https://www.elastic.co/guide/en/elasticsearch/reference/current/near-real-time.html) occurs between these requests, the order of your results may change, causing inconsistent results across pages. To prevent this, you can create a [point in time (PIT)](https://www.elastic.co/guide/en/elasticsearch/reference/current/point-in-time-api.html) to preserve the current index state over your searches.

I think this api is obviously useful for large searches, but I think we could also leverage it for our smaller range pagination on search results to avoid the duplicated search results.

scripts/identify-ids-by-query.js
```diff
@@ -0,0 +1,189 @@
...
+        default: '100'
+      },
+      stripprefix: {
+        type: 'boolean',
+        default: false
+      }
+    }
+  })
+
+  castArgsToInts(argv, ['batchSize', 'limit', 'offset'])
```
charmingduchess, line 84: I think this line shouldn't be here. These arguments are not specified in the preamble.

scripts/identify-ids-by-query.js
```diff
@@ -0,0 +1,189 @@
...
+    options: {
+      envfile: {
+        type: 'string',
+        default: './config/qa.env'
+      },
+      from: {
+        type: 'string',
+        default: '0'
+      },
+      outfile: { type: 'string' },
```
charmingduchess, line 69: the preamble says this field has a default, but there is not one specified here.

scripts/identify-ids-by-query.js
nonword: Good catch. We do want to support `--limit`, so keeping that one.

charmingduchess:
Can you update this script name to include ES?

charmingduchess (approved):

PR #110, opened 2025-02-12
[SCC-4363] Add field 520b to index

yossariano:
* Ticket: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4363
* This seems like it could add a huge chunk of searchable text to the index- is that OK? Will it result in undue noise? We'll see!


charmingduchess (approved):
I think this bib has particularly verbose metadata because it's an archive-like record. I don't think most bibs have this kind of detail. But for bibs like this, this field is currently searchable in legacy, and it makes sense. We also have been pretty liberal in updating these indexed values, basically adding anything the librarians complain about not being able to find. While it's maybe not the most systematic way to go about things, I think it gets us closer to that parity with legacy, and generally keeps the librarians happy. I've found that in general librarians actually prefer the noise in search results to disconcerting quiet... 

That said, it does seem like for this bib, most of the key words in the summary are also included in subject heading fields. It feels redundant. We're constantly navigating nuances between official cataloguing standards that we've based a lot of our work on, and the actual facts of how the different research branches actually have catalogued their zillions of records, which in turn is influenced by the specific legacy UI. Librarians have adopted many different, localized, non standard practices in order to get things to display in specific ways in legacy.

nonword (approved):

PR #111, opened 2025-02-28
add formatId to indexing process

charmingduchess:


lib/es-models/bib.js
```diff
@@ -619,6 +619,13 @@ class EsBib extends EsBase {
     else return null
   }
 
+  formatId () {
+    if (this.bib.isPartnerRecord()) {
+      return this.recordTypeId()
+    }
+    return this.bib.fixed('materialType') || null
```
nonword, line 626: I believe the key will be `Material Type`

lib/es-models/bib.js
```diff
@@ -619,6 +619,13 @@ class EsBib extends EsBase {
     else return null
   }
 
+  formatId () {
+    if (this.bib.isPartnerRecord()) {
+      return this.recordTypeId()
+    }
+    return this.bib.fixed('Material type') || null
```
nonword, line 626: 'Material Type' (capital t). Also, [looks like `fixed`](https://github.com/NYPL/research-catalog-indexer/blob/6e45f68862f90894b5881f64e39f71e1a8c8a307/lib/sierra-models/base.js#L15) returns the value without stripping it, so we may want to strip it since I see values like `"e  "`

nonword (approved):

PR #119, opened 2025-05-29
SCC-3727 update what fields are indexed as subjectLiteral

charmingduchess:
- incorporates the new browse-term npm package to build out `subjectLiteral` on the es bib model. 
- via npm package, now there will be more subfields being indexed than previously were listed in the subject literal bib mapping
- add 690 field (local subjects) to bib-mapping json
- fix bug in bulk index where last id was not indexing


lib/es-models/bib.js
```diff
@@ -10,14 +10,15 @@ const { titleSortTransform } = require('../utils/title-sort-transform')
...
 const { unique, removeTrailingElementsMatching, sortByAsyncSortKey } = require('../utils')
-const { subjectLiteralFromSubfieldMap } = require('../utils/subject-formatter')
 const {
   normalizeAuthorName,
   withoutDates,
   nameIsRedundant,
   trimTrailingPeriod
 } = require('../utils/author-names')
 
+const { firstSubfields, lastSubfields, subjectLiteralFromSubfieldMap } = require('@nypl/browse-term')
```
nonword, line 20: Looks like the npmjs hosted package is ahead of the git repo?

charmingduchess: Just updated.

charmingduchess:
@nonword the code currently references some local changes to the npm module that I don't want to publish until you've approved this approach. The approach being, to pass along an `includeMarc` flag for subject literal varfield march objects, which allows filtering to happen as usual, but also for the marc to remain accessible to the Subject class further down the line for label processing.

lib/sierra-models/base.js
```diff
@@ -71,13 +71,7 @@ class SierraBase {
...
   }
 
-  // This function takes a marc tag, and returns an array of primary values with
-  //   relevant parallel values attached and or orphaned parallels relevant to
-  //   given marc tag
-  _fieldsWithParallelsAndOrPrimaries (marc) {
-    // _varFieldByMarcTag returns array of raw marc fields filtered by marc tag:
-    //  [{fieldTag: 'a', marcTag: '123', content: 'the thing'}]
-    const primaryRawFields = this._varFieldByMarcTag(marc)
+  _buildParallelsLookup (marc) {
```
charmingduchess, line 74: this was extracted for an approach I did not end up using, but I neglected to reintegrate it.

lib/sierra-models/base.js
```diff
@@ -253,17 +257,17 @@ class SierraBase {
    *
    * @return {VarFieldMatch[]}
    */
-  varFieldsMulti (marcObjects) {
+  varFieldsMulti (marcObjects, includeMarc = false) {
```
charmingduchess, line 260: I was debating between adding `includeMarc` as a mapping option on the bib mapping, let me know if there is maybe a better place to introduce this flag.

nonword: As discussed, this makes sense to me now.

lib/es-models/bib.js
```diff
@@ -1004,25 +1005,15 @@ class EsBib extends EsBase {
...
-    })
-    // Remove empty trailing subjects:
-    values = removeTrailingElementsMatching(values, (s) => s === '')
+    const marcTags = BibMappings.get('subjectLiteral', this.bib)
+    const subjectLiterals = this.bib.varFieldsMulti(marcTags, true).map((subjectVarfieldObject) => {
+      const marc = primary ? subjectVarfieldObject.marc : subjectVarfieldObject.parallel.marc
+      if (!marc) return null
+      const subject = new Subject({ varFields: [marc] }, primary ? 'd' : 'y')
+      return subject.preferredTerm.label
+    }).filter(x => x)
```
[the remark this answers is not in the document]
nonword: As discussed, I retract the concerns about having to maintain the marc tag config in two places given that it's only likely to live in RCI. Long term we _may_ want to give more of this over to the module, but I don't see a compelling argument to do that now. The interface you're proposing is good.

nonword (approved):

PR #122, opened 2025-06-11
Update bulk-index to support bulk-indexing all

nonword:
The hasMarc and nyplSource filters are now optional


scripts/bulk-index.js
```diff
@@ -613,7 +623,7 @@ const run = async () => {
   if (
     argv.bibId ||
     argv.itemId ||
-    (argv.type && argv.hasMarc)
+    (argv.type)
```
charmingduchess, line 626: Does this leave the script open to an update on all of a given type? I feel like we should encourage a practice of narrowing down slightly the scope... maybe a check like `argv.type && (argv.hasMarc || argv.nyplSource)`

charmingduchess (changes requested):

scripts/bulk-index.js
nonword: Will implement

nonword:
I updated the arg validation to require either nyplSource or hasMarc (although looks like currently nyplSource defaults to 'sierra-nypl', so it's already effectively required). And I updated the USEAGE note to acknowledge the new useage

charmingduchess (dismissed):

PR #144, opened 2025-07-25
[SCC-4795] Optionally support bib_v2 and date ranges

yossariano:
* This adds a few arguments to the bulkIndex script, namely:
  * `table` so we can specify `bib_v2` or `bib` and compare performance
  * `fromDate` and `toDate` as I was trying to get big amounts of bibs to validate with. How are we normally using this to do (for instance) a full reindex? 

#### Testing
* In my testing- and keep in mind I was using `limit` so I wasn't doing a massive amount of bibs- I don't really see a performance difference between `bib` and `bib_v2` with the updated indices. Would like to know if folks expect us to use this differently in some way or how we were seeing the most slowdown before.


charmingduchess:
Paul recently updated the bulk index script to accept an nypl source, to facilitate doing a reindex of everything. I'm not sure that we've ever actually triggered a full reindex before from this script.

nonword (approved):

nonword:
I started working on something overlapping, so proposing [these changes](https://github.com/NYPL/research-catalog-indexer/pull/147) - I think we can skip the inner-select in general when not doing `hasMarc` queries.

yossariano:
> I started working on something overlapping, so proposing [these changes](https://github.com/NYPL/research-catalog-indexer/pull/147) - I think we can skip the inner-select in general when not doing `hasMarc` queries.

Cool, that looks good to me. I can discard this PR in favor of that one.

yossariano:
Closing in favor of https://github.com/NYPL/research-catalog-indexer/pull/147

PR #146, opened 2025-07-29
Fix bulk-index: No inner-select when not needed

nonword:
Removing the inner-select for straightforward (no varfield matching) queries should improve performance.

Also adds support for `--tablename T` override so we can easily test `bib_v2`


nonword:
Closing in favor of https://github.com/NYPL/research-catalog-indexer/pull/144 and https://github.com/NYPL/research-catalog-indexer/pull/147

PR #147, opened 2025-07-29
Slightly different check for skipping the inner-select

nonword:
The thing that should trigger the inner-select is querying by repeatable fields (i.e. `hasMarc`). For everything else, we don't need the inner-select


yossariano (approved):
Nice- from my testing this looks like a big speedup.

PR #148, opened 2025-08-05
Scc 3323/subject order

charmingduchess:
Add sorting on the `varFieldsMulti` method, based on the index of a given varfield in the varfields array returned from BibService, to respect the order that varfields are catalogged in. This change was added to address subjects, but should enforce sorting on all varfield multi fields.
 - featuring some finicky overriding biz due to orphan parallels

### DIFF
--- a/lib/sierra-models/base.js
+++ b/lib/sierra-models/base.js
@@ -269,7 +269,7 @@ class SierraBase {
       .sort((a, b) => {
         // This is for the edgiest of cases where 880 fields appear first in a varfield array, which will probably never happen.
         const isOrphanedParallel = (varfield) => !varfield.value
-        // if either a or be is exclusively orphaned..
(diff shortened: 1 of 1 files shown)


lib/sierra-models/base.js
```diff
@@ -269,8 +269,12 @@ class SierraBase {
       .sort((a, b) => {
         // This is for the edgiest of cases where 880 fields appear first in a varfield array, which will probably never happen.
         const isOrphanedParallel = (varfield) => !varfield.value
-        if (isOrphanedParallel(a)) a.varFieldIndex = 1000
-        if (isOrphanedParallel(b)) b.varFieldIndex = 1000
+        // if either a or be is exclusively orphaned..
```
nonword, line 272: Whoops. `be` => `b`

nonword (dismissed):

PR #151, opened 2025-08-08
add source id to subject event

charmingduchess:
- adding bib ids for debugging and to be able to delete/reindex by bibId

### DIFF
--- a/lib/browse-terms/index.js
+++ b/lib/browse-terms/index.js
@@ -45,7 +45,7 @@ const getPrimaryAndParallelLabels = (subjectVarfieldObject) => {
 
 const getSubjectModels = (sierraRecord) => {
   const marcTags = BibMappings.get('subjectLiteral', sierraRecord)
-  return sierraRecord.varFieldsMulti(marcTags, true).map(getPrimaryAndParallelLabels)
+  return sierraRecord.varFieldsMulti(marcTags, true).map(getPrimaryAndParallelLabels).map((subject) => ({ ...subject, sourceId: sierraRecord.id }))
 }
 /**
  *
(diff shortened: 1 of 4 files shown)


lib/browse-terms/index.js
```diff
@@ -45,7 +45,7 @@ const getPrimaryAndParallelLabels = (subjectVarfieldObject) => {
 
 const getSubjectModels = (sierraRecord) => {
   const marcTags = BibMappings.get('subjectLiteral', sierraRecord)
-  return sierraRecord.varFieldsMulti(marcTags, true).map(getPrimaryAndParallelLabels)
+  return sierraRecord.varFieldsMulti(marcTags, true).map(getPrimaryAndParallelLabels).map((subject) => ({ ...subject, sourceId: sierraRecord.id }))
```
yossariano, line 48: If we eventually get the same subject from an authority record (whether because it was indexed via a bib first or just added later) will that override the sourceId to the authorityRecordId? And what about the inverse?

yossariano (approved):
Looks good, one minor question about how this interplays with authority record ids.

lib/browse-terms/index.js
charmingduchess: it's only run on bib events, and the processing for authority events should mean that they always overwrite a bib or authority subject already in the index with the same preferred term

charmingduchess: Sorry i think i answered your question backwards - an authority record will overwrite sourceId of an existing bib subject, but not the other way around. Once a preferred term is in the index, only an authority update will overwrite any of the properties.

charmingduchess: we have source: bib/authority on the ES documents, and the sourceIds are also distinguishable by prefix (or lack thereof)

PR #152, opened 2025-08-11
rm extra term value from bib event

charmingduchess:


charmingduchess:
Actually going to keep this event type and update browse term indexer to reference that instead of the SQS stream name

charmingduchess:
Actually, on second thought, I am going to go ahead with this.

PR #165, opened 2025-09-16
Actually skip mget when no idsToFetch

charmingduchess:


danamansana (approved):
One small comment but makes sense to me

PR #168, opened 2025-09-16
Noref/handle null dates

danamansana:
- Adds a check for inconsistent date ranges and rejects them. We may come back and do something more specific later
- Adds a test


lib/utils/es-ranges.js
```diff
@@ -80,6 +80,9 @@ const generateDateRangeFromYears = (first, second, rawMarc, type) => {
   const secondRoundedUp = roundDateUp(second)
   if (!rawMarc || !rawMarc.length) { return null }
   if (!firstRoundedDown) { return null }
+  if (parseInt(firstRoundedDown) >= (parseInt(secondRoundedUp) + 1)) {
```
yossariano, line 83: Why do `>= x + 1` instead of `> x`?

danamansana: Good point. I've changed that

charmingduchess (approved):

PR #170, opened 2025-09-19
Add some logging around dates

danamansana:
- Log ids of bibs with inconsistent date ranges
- Log ids of bibs missing dates (I am on the fence about this one -- probably a lot of these are mistakes but also a lot are intentional, and we have other ways at getting at this information. What do you think?)

### DIFF
--- a/lib/es-models/bib.js
+++ b/lib/es-models/bib.js
@@ -17,7 +17,8 @@ const {
   trimTrailingPeriod
 } = require('../utils/author-names')
 const {
-  generateDateRangeFromYears
+  generateDateRangeFromYears,
+  InconsistentDateRange
 } = require('../utils/es-ranges')
 const logger = require('../logger')
 
@@ -227,7 +228,6 @@ class EsBib extends EsBase {
   dates () {
     const dates = []
     const rawMarc = this.bib.varField('008')
-    const id = this.bib.id
     if (!rawMarc || !rawMarc.length) { return [] }
     const type = this.bib.varFieldSegment('008', [6, 6])?.trim()
     const first = this.bib.varFieldSegment('008', [7, 10])?.trim()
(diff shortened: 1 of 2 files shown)


lib/es-models/bib.js
```diff
@@ -256,7 +258,11 @@ class EsBib extends EsBase {
         tag: type
       })
     }
-    return dates.filter(x => x)
+    const filteredDates = dates.filter(x => x)
+    if (!filteredDates.length) {
+      logger.warn(`Bib missing dates: ${id}`)
+    }
```
nonword, line 264: I think I agree with your concern that this is going to generate a lot of noise.

lib/es-models/bib.js
```diff
@@ -226,14 +227,15 @@ class EsBib extends EsBase {
   dates () {
     const dates = []
     const rawMarc = this.bib.varField('008')
+    const id = this.bib.id
```
nonword, line 230: Nit: Given changes suggested below, if this alias has very little or singular use, maybe lets not create it?

lib/es-models/bib.js
```diff
@@ -233,7 +235,16 @@ class EsBib extends EsBase {
     // these types indicate date ranges. The 008 field is interpreted as a
     // single range defined by two YYYY dates
     if (['c', 'd', 'i', 'k', 'u'].includes(type)) {
-      dates.push(generateDateRangeFromYears(first, second, rawMarc, type))
+      try {
+        dates.push(generateDateRangeFromYears(first, second, rawMarc, type))
+      } catch (e) {
+        if (e instanceof InconsistentDateRange) {
+          logger.warn(`Inconsistent date range ${this.bib.id}. Marc: ${rawMarc}`)
```
yossariano, line 242: extreme nitpick: We should adjust this log line so it indicates that the string after "range" is a bib id. Something like

```
Inconsistent date range. bibId: X Marc: Y
```

or better yet, use structured logging for this. Looks like we already have json logging enabled in this package [here](https://github.com/NYPL/research-catalog-indexer/blob/main/lib/logger.js#L5) so we could pass in a js object and get 

```
logger.warn('Inconsistent date range.', { bibId: 123 });
```

and it should log something like

```
{"message": "Inconsistent date range.", "bibId": 123}
```

Though, in light of us not using that anywhere else in this package yet, I would say it's not super important. But it would make parsing bibIds out of logs a breeze.

yossariano (approved):

PR #184, opened 2025-10-03
Noref/skip blank dates

danamansana:
A previous PR added some extra warning for invalid dates. This resulted in a lot of noisy extraneous warnings for multiple-date types when the second date is blank. This PR fixes this by skipping the second date if it is blank

### DIFF
--- a/lib/es-models/bib.js
+++ b/lib/es-models/bib.js
@@ -273,16 +273,7 @@ class EsBib extends EsBase {
       // these types indicate date ranges. The 008 field is interpreted as a
       // single range defined by two YYYY dates
       if (['c', 'd', 'i', 'k', 'm', 'u'].includes(type)) {
-        try {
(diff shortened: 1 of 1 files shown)


lib/es-models/bib.js
```diff
@@ -288,7 +288,9 @@ class EsBib extends EsBase {
       // separate YYYY dates, each of which is indexed as a 1-year range
       if (['p', 'q', 'r', 's', 't'].includes(type)) {
         dates.push(generateDateRangeFromYears(first, first, rawMarc, type))
-        dates.push(generateDateRangeFromYears(second, second, rawMarc, type))
+        if (!second.match(/^[u\s|]{4}$/)) {
```
yossariano, line 291: Could you add a comment explaining / giving an example of what this regex is checking for? Do we get date strings like `uuuu`, `u uu` etc or am I misreading this?

yossariano (approved):

PR #285, opened 2026-03-13
update env to use aliases

charmingduchess:
Reference aliases instead of explicit index names 
Tested by reindexing b10833141 in qa and prod, verifying that record has recent updated date and bib looks good on RC.


yossariano (approved):

nonword (commented):
When changing a field that requires updates in RCI and discovery-api, I assume the procedure is:
 - create new index with different schema
 - locally test writing and reading from the new index (by direct reference, not alias)
 - when satisfied, create PRs for both RCI and discovery-api
 - after both approved, merge both simultaneously
 - update qa alias
 - accept that there may be a brief period when RCI-qa is erroring because of the unexpected mapping change and/or discovery-api is quietly failing some searches because it's matching on legacy fields

That right?

PR #286, opened 2026-03-17
Update bib sort fields to strip leading articles

yossariano:
* Ticket: https://newyorkpubliclibrary.atlassian.net/browse/SCC-4597
* This updates title and contributor sorts to strip out leading articles ("a", "an" and "the") for sorting purposes.
* Note that I haven't actually updated the index settings / mappings for this- I think we'll need to make a new index and reindex into it to change the normalizer on these fields and I would like to get eyes on the proposed change before doing so.

### DIFF
--- a/lib/elastic-search/index-schema.js
+++ b/lib/elastic-search/index-schema.js
@@ -17,6 +17,12 @@ const propertyTemplates = {
   exactString: {
     type: 'keyword'
   },
+  // Used in sorting- this will strip out leading articles ("the", "a", and "an")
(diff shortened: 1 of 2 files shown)


nonword (approved):

charmingduchess (commented):
We ought to account for articles in other languages, too. The proper way to do this would be to update the _sort values to respect the second indicator position described [here](https://www.loc.gov/marc/bibliographic/bdapndxf.html), and not index those articles at all.

PR #295, opened 2026-03-26
Qa -> production

danamansana:
I'd like to merge this to prod for SCC-5160 (which is about allowing multiple call numbers per bib) but looks like there are a lot of other changes here. Can people take a look and see if their changes are ready to go?


nonword (approved):
My changes are good to go

yossariano (approved):
Contributor browse stuff should be good to go

PR #313, opened 2026-04-21
Add lambda alarm configured via terraform

yossariano:
* This shows how to create a lambda alarm via terraform
* One big difference between this approach and our current approach is this sends the alarm to a generic `research-catalog-team-alarms-${environment}` topic rather than individual service topics that has been preconfigured to forward alarm messages to slack.

### DIFF
--- a/provisioning/base/resources.tf
+++ b/provisioning/base/resources.tf
@@ -63,3 +63,27 @@ resource "aws_lambda_function" "lambda_instance" {
   }
   
 }
+
+data "aws_sns_topic" "rc_alarms" {
+  name = "research-catalog-team-alarms-${var.environment}"
+}
+
+resource "aws_cloudwatch_metric_alarm" "lambda_errors" {
+  alarm_name          = "lambda-errors-${aws_lambda_function.lambda_instance.function_name}"
+  comparison_operator = "GreaterThanThreshold"
+  evaluation_periods  = 1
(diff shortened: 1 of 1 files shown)


provisioning/base/resources.tf
```diff
@@ -63,3 +63,27 @@ resource "aws_lambda_function" "lambda_instance" {
...
+resource "aws_cloudwatch_metric_alarm" "lambda_errors" {
+  alarm_name          = "lambda-errors-${aws_lambda_function.lambda_instance.function_name}"
+  comparison_operator = "GreaterThanThreshold"
+  evaluation_periods  = 1
+  metric_name         = "Errors"
+  namespace           = "AWS/Lambda"
+  period              = 300
+  statistic           = "Sum"
+  threshold           = 1
+  alarm_description   = "Lambda function ${aws_lambda_function.lambda_instance.function_name} has more than 1 error in 5 minutes"
```
charmingduchess, line 80: ie alarm is triggered on the second error? does this continue to send alarms on 3rd, 4th, etc?

charmingduchess (approved):

provisioning/base/resources.tf
yossariano: It will just trigger on the first error, and will stay "in alarm" for subsequent ones without additional notification. If it goes green again, the next time it turns red it'll send a new notification.

PR #317, opened 2026-04-27
update index schema

charmingduchess:
This pr is mostly about creating WIP script for creating new indexes based on current ones (see this [ticket](https://newyorkpubliclibrary.atlassian.net/browse/SCC-5365) for proposed future work).

but also updates `*series*` fields with new property template to reflect new index


lib/elastic-search/index-schema.js
7emansell, line 1: are the changes here still relevant?

scripts/copy-qa-index-to-prod.js
7emansell, line 1: @charmingduchess

lib/elastic-search/index-schema.js
charmingduchess: yes... i mostly had them there to save the mappings i was using. do you think they should be removed?

PR #323, opened 2026-05-15
bulk shell script

charmingduchess:
New approach to reingesting:
Fetch allll ids regardless of isResearch or owner 
Split into 1000-id length files
Invoke bulk-index per file

Is there maybe a way to run these in parallel? I think currently they are running in series. Do the pools that the PG client uses only exist in the node environment or are they actual pg entities that we would have access to outside of the process?

A version of this script has successfully run a bulk-update in prod in around 28 hours. This exact script has been successfully tested in qa with a `WHERE id = 10833141` clause added to the SQL query.


scripts/bulk_index_chunk_runner.sh
```diff
@@ -0,0 +1,34 @@
...
+# $@  -  rest of flags passed to script
+
+SCRIPT_DIR=.
+mkdir -p $SCRIPT_DIR/tmp/
+BIB_IDS_DIR=$SCRIPT_DIR/tmp/bib_ids/
+ERRORS_DIR=$SCRIPT_DIR/tmp/chunk_run_errors/
+mkdir -p $BIB_IDS_DIR
+mkdir -p $ERRORS_DIR
+
+set -a; source config/"$1"-bulk-index.env; set +a
```
danamansana: what is set -a set +a?

charmingduchess: I'm not totally sure, but it is likely superfluous. Paul sent me this but then amended his recommendation

nonword: Those enable automatic export for variables defined in the sourced file, but I don't believe they're necessary here. You can just `source config/"$1"-bulk-index.env` and the env vars should be available because the behavior of `source` is to run the commands found in the target file _within the current shell_. Hence any command of the form `VAR=val` should cause `$VAR` to be set just as if you ran it in the the current script file. It looks like this is true of both zsh & bash

scripts/bulk_index_chunk_runner.sh
```diff
@@ -0,0 +1,38 @@
...
+mkdir -p $SCRIPT_DIR/tmp/
+BIB_IDS_DIR=$SCRIPT_DIR/tmp/bib_ids/
+ERRORS_DIR=$SCRIPT_DIR/tmp/chunk_run_errors/
+PROCESSED_IDS_DIR=$SCRIPT_DIR/tmp/processed/
+mkdir -p $BIB_IDS_DIR
+mkdir -p $ERRORS_DIR
+mkdir -p $PROCESSED_IDS_DIR
+
+if [ "$2" == "true" ]; then
+  set -a; source config/"$1"-bulk-index.env; set +a
```
nonword: This `set -a` shouldn't be necssary

scripts/bulk_index_chunk_runner.sh
```diff
@@ -0,0 +1,38 @@
...
+  DECRYPTED_PW=$(kms-util decrypt $BIB_SERVICE_DB_PW)
+  DECRYPTED_HOST=$(kms-util decrypt $BIB_SERVICE_DB_HOST)
+  DECRYPTED_USER=$(kms-util decrypt $BIB_SERVICE_DB_USER)
+	echo Fetching all bib ids from $1 bib database
```
nonword, line 22: Indentation disagreement

scripts/bulk_index_chunk_runner.sh
```diff
@@ -0,0 +1,38 @@
...
+SCRIPT_DIR=.
+mkdir -p $SCRIPT_DIR/tmp/
+BIB_IDS_DIR=$SCRIPT_DIR/tmp/bib_ids/
+ERRORS_DIR=$SCRIPT_DIR/tmp/chunk_run_errors/
+PROCESSED_IDS_DIR=$SCRIPT_DIR/tmp/processed/
+mkdir -p $BIB_IDS_DIR
+mkdir -p $ERRORS_DIR
+mkdir -p $PROCESSED_IDS_DIR
+
+if [ "$2" == "true" ]; then
```
nonword: Suggest initializing this as `FETCH_IDS=$2` at the top of the script so that this conditional can read `if [ "$FETCH_IDS" == "true" ];`

scripts/bulk_index_chunk_runner.sh
```diff
@@ -0,0 +1,38 @@
...
+  set -a; source config/"$1"-bulk-index.env; set +a
+  DECRYPTED_PW=$(kms-util decrypt $BIB_SERVICE_DB_PW)
+  DECRYPTED_HOST=$(kms-util decrypt $BIB_SERVICE_DB_HOST)
+  DECRYPTED_USER=$(kms-util decrypt $BIB_SERVICE_DB_USER)
+	echo Fetching all bib ids from $1 bib database
+	psql postgresql://$DECRYPTED_USER:$DECRYPTED_PW@$DECRYPTED_HOST/bib_service_production -c "\\COPY (SELECT id, nypl_source FROM bib) TO $SCRIPT_DIR/tmp/all_bib_ids.csv WITH CSV DELIMITER ',' HEADER;" 
+
+	echo Splitting `wc -l $SCRIPT_DIR/tmp/all_bib_ids.csv` into files of 1000 ids
+split -l 1000 -a 5 -d $SCRIPT_DIR/tmp/all_bib_ids.csv $BIB_IDS_DIR/bib_ids_
```
nonword: Indent?

scripts/bulk_index_chunk_runner.sh
```diff
@@ -0,0 +1,40 @@
...
+  DECRYPTED_HOST=$(kms-util decrypt $BIB_SERVICE_DB_HOST)
+  DECRYPTED_USER=$(kms-util decrypt $BIB_SERVICE_DB_USER)
+	echo Fetching all bib ids from $1 bib database
+	psql postgresql://$DECRYPTED_USER:$DECRYPTED_PW@$DECRYPTED_HOST/bib_service_production -c "\\COPY (SELECT id, nypl_source FROM bib) TO $SCRIPT_DIR/tmp/all_bib_ids.csv WITH CSV DELIMITER ',' HEADER;" 
+
+	echo Splitting `wc -l $SCRIPT_DIR/tmp/all_bib_ids.csv` into files of 1000 ids
+  split -l 1000 -a 5 -d $SCRIPT_DIR/tmp/all_bib_ids.csv $BIB_IDS_DIR/bib_ids_
+
+	echo Deleting original $SCRIPT_DIR/tmp/all_bib_ids.csv file
+	rm $SCRIPT_DIR/tmp/all_bib_ids.csv
```
nonword, line 29: Let's apply uniform indentation throughout. Looks like some 2-width tabs, some 4

scripts/bulk_index_chunk_runner.sh
nonword: Now that you have `FETCH_IDS` established as a var, use it here

```
if [ "$FETCH_IDS" == "true" ]; then
```

scripts/bulk_index_chunk_runner.sh
```diff
@@ -0,0 +1,40 @@
...
+ERRORS_DIR=$SCRIPT_DIR/tmp/chunk_run_errors/
+PROCESSED_IDS_DIR=$SCRIPT_DIR/tmp/processed/
+FETCH_IDS=$2
+
+mkdir -p $BIB_IDS_DIR
+mkdir -p $ERRORS_DIR
+mkdir -p $PROCESSED_IDS_DIR
+
+if [ "$2" == "true" ]; then
+  source config/"$1"-bulk-index.env; set +a
```
nonword: I meant the `set` calls are unneeded in general. Free to remove the trailing `set +a` call

nonword (approved):

PR #332, opened 2026-05-29
mapping updates

charmingduchess:
- fix initialize index script to await cli input
- add 3 missing mappings determined by discovery-api mapping-check.js
- update qa index


scripts/initialize-index.js
```diff
@@ -29,35 +29,41 @@ exports.run = async (options = {}) => {
   const exists = (await client.indices.exists({ index: options.index })).body
 
   if (exists) {
-    die(`Index ${options.index} exists. Exiting`)
-  }
-
-  console.log(`Initializing ${options.index}`)
+    console.log(`Index ${options.index} exists.`)
```
7emansell, line 32: should this still stop the script?

scripts/initialize-index.js
```diff
@@ -29,35 +29,41 @@ exports.run = async (options = {}) => {
...
+
+const optionallyCopyContentsToNewIndex = async (newIndexName) => {
+  const client = await esClient.client()
   const reindexRl = readline.createInterface({
     input: process.stdin,
     output: process.stdout
   })
-  const oldIndex = process.env.RESOURCES_INDEX
-  reindexRl.question(`copy contents of ${oldIndex} to ${options.index}? Only "yes" will trigger copy... `, async answer => {
+  const oldIndex = process.env.ELASTIC_RESOURCES_INDEX_NAME
```
7emansell, line 56: is this name always the same as whatever value is in the `config/*.env`?

7emansell (commented):
Looks good just questions

scripts/initialize-index.js
```diff
@@ -29,35 +29,41 @@ exports.run = async (options = {}) => {
...
+          dynamic: 'strict',
+          properties: schema()
+        }
       }
-    }
-  })
+    })
+    console.log(`Index ${options.index} initialized.`)
+  }
+  await optionallyCopyContentsToNewIndex(options.index)
```
nonword, line 47: Yeah per Emma's comment above, probably want to move this into the `else` block above so that we don't inadvertently start a reindex into an existing index?

charmingduchess: That was on purpose, but it doesn't have to be like that

7emansell (approved):

nonword (approved):

PR #334, opened 2026-06-09
prod merge

charmingduchess:
If prod didn't have these changes, do we need to rewind the indexer again?


7emansell (approved):
I don't think prod needs to be rewound (at least not again, i think it was rewound to 5/26 post the most recent index replacement) since the only significant changes would be in `index-schema` and those made it in

config/production-browse-ingest.env
7emansell: this is in production github is wack

nonword (approved):
All looks good to me.

lib/es-models/item.js
```diff
@@ -65,6 +65,23 @@ class EsItem extends BaseItemMixin(EsBase) {
...
+      const entry = Object.values(nyplCore.collectionMapping()).find((c) =>
+        c.holdingLocations.includes(holdingId)
+      )
+      if (entry) {
+        return [entry.code]
+      }
+    }
+    return null
+  }
+
```
nonword, line 96: This change means we'll start indexing item.collectionId. I imagine we're not relying on that field being populated currently in production.

lib/es-models/item.js
```diff
@@ -200,6 +217,11 @@ class EsItem extends BaseItemMixin(EsBase) {
       literal = item._recapCustomerCode
     } else if (!item.isNyplRecord()) {
       literal = item.varField('900', ['b'])[0]?.value
+
+      // To ensure uniqueness of code, prefix alternate depository codes:
+      if (item._depository() && item._depository() !== 'recap') {
+        literal = `${item._depository()}:${literal}`
+      }
```
nonword, line 236: This is mine and is good to go.

nonword (approved):

PR #335, opened 2026-06-11
SCC-5434: Move `collection` onto items

7emansell:
context: with status messaging display updates AND lists, RC now needs to know collection ID mapped to item rather than just all the collections a bib is in. No change to bib unit tests since the `collection` field on the bib should be identical, it's just collecting the IDs from its items now

### DIFF
--- a/lib/es-models/bib.js
+++ b/lib/es-models/bib.js
@@ -151,21 +151,13 @@ class EsBib extends EsBase {
       )
       return match ? [match.code] : []
     }
-    // Translate item holdingLocations into plain ids:
-    const locationCodes = items.flatMap((item) => {
-      const loc = item.holdingLocation()
-      if (loc && loc.length) {
-        return loc[0].id.split(':').pop() // extract plain id
+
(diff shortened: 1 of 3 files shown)


charmingduchess (approved):

lib/es-models/bib.js
```diff
@@ -151,22 +151,14 @@ class EsBib extends EsBase {
...
     }
-    // Translate item holdingLocations into plain ids:
-    const locationCodes = items.flatMap((item) => {
-      const loc = item.holdingLocation()
-      if (loc && loc.length) {
-        return loc[0].id.split(':').pop() // extract plain id
+
+    // Collect collection/division ids from items:
+    const collectionCodes = items.flatMap((item) => {
+      if (item.collection) {
```
nonword, line 157: Should this be `item.collection()`?

nonword: OH I see. This is intentionally checking if the function exists. Got it.

nonword (approved):

PR #337, opened 2026-06-23
Add automatic retrying for fetch in utils

danamansana:
This was ticketed but not sure we every decided we definitely want to do this so feel free to weigh in.
This PR adds an automatic retry in case the fetch in `nypl-source-mapper` fails. It is intended to help quiet the alarms

### DIFF
--- a/lib/utils/nypl-source-mapper.js
+++ b/lib/utils/nypl-source-mapper.js
@@ -56,15 +56,24 @@ class NyplSourceMapper {
 const createInstance = async () => {
   const sourceMappingUrl = `https://raw.githubusercontent.com/NYPL/nypl-core/${process.env.NYPL_CORE_VERSION || 'master'}/mappings/recap-discovery/nypl-source-mapping.json`
 
-  // Retrieve json file:
-  const resp = await fetch(sourceMappingUrl)
-    .catch((e) => {
-      throw new Error(`Error retrieving ${sourceMappingUrl}: ${e}`)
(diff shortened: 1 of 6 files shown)


lib/utils/nypl-source-mapper.js
```diff
@@ -56,15 +56,24 @@ class NyplSourceMapper {
...
+    } catch (err) {
+      if (attempt === 1) {
+        logger.warn(`First fetch attempt failed for ${sourceMappingUrl} (${err.message}). Retrying...`)
+      } else {
+        const msg = resp && !resp.ok
+          ? `Error retrieving ${sourceMappingUrl} - got status ${resp.status}`
+          : `Error retrieving ${sourceMappingUrl}: ${err.message}`
+        throw new Error(msg)
+      }
+    }
```
nonword, line 66: Suggesting we move [this function](https://github.com/NYPL/research-catalog-indexer/blob/f4c408b4903ea167f95d8653394579de87117a6c/scripts/utils.js#L367-L403) into a more central utils file and use it. The logic of retrying is abstractable. Also gives us configurable numbers of retries and backoff.

danamansana: Good idea. I've created a new `retry` util, and updated the app to use it in all places the previous retry was being used, as well as replacing the retry logic in the `nypl-source-mapper` with the new util.

nonword (approved):

PR #338, opened 2026-06-24
Update 'Copy No.' parsing to ignore vals < 2

nonword:
https://newyorkpubliclibrary.atlassian.net/browse/SCC-5453


danamansana (approved):
Good fix!

PR #342, opened 2026-07-15
Clean up some mappings for clarity

nonword:
This just renames some things for consistency. Should be no change to indexing rules.


charmingduchess (dismissed):
Fine by me, though I'm curious about the callnumber/shelfmark distinction.

PR #346, opened 2026-07-27
Scc 5497/browse term config

charmingduchess:
- Update all bulk index config to write browse term updates to file, not to the live sqs stream. This is because we only want the stream of organic updates to write to SQS in the deployed RCI. Local invocations of the bulk index should never trigger the live BTI.
- create partial index config files to facilitate bib only updates. intended to be used as follows:
    - `node scripts/bulk-index.js --envfile config/qa-partial-index.env --properties series,uniformTitle,subject --type bib --csv {path to csv}`

Note these will require manual intervention at the end of the index to run the browse term updates through a local version of the Browse term indexer. However, this is only necessary when updates are being made to browse term fields, ie on a total reindex and/or when we are changing how we index browse terms (author, contributor, subject).

Also, github decided that I renamed and edited the browse ingest files, but i in fact deleted and created new files with very similar contents. the browse ingest files are no longer needed since we changed how much we query ES during the browse term parsing, as well as the new flow of writing browse terms to file.


config/production-bulk-index.env
```diff
@@ -38,4 +35,7 @@ SIERRA_SECRET=AQECAHh7ea2tyZ6phZgT4B9BDKwguhlFtRC6hgt+7HbmeFsrsgAAAIcwgYQGCSqGSI
 SKIP_DOC_PROCESSED_STREAM=true
 LOG_LEVEL=info
 EMIT_BROWSE_TERMS=true
+BROWSE_TERM_OUTPUT_DIRECTORY=./browse-term-output-prod.csv
```
danamansana: technically this is a file not a directory

danamansana (dismissed):
looks good just one small comment on naming

config/production-bulk-index.env
charmingduchess: good catch. should be a directory

lib/browse-terms/index.js
```diff
@@ -51,18 +56,34 @@ const emitBrowseTerms = async (esDocuments, termType) => {
       const batchedEvents = convertBatchToEventForLocalBti(bulkBrowseTerms)
       const timestamp = Date.now()
       const eventFile = `${process.env.BROWSE_TERM_OUTPUT_DIRECTORY}/${termType}-${timestamp}.json`
-      logger.info(`Emitting ${uniqueTerms.length} ${termType} browse terms across ${bulkBrowseTerms.length} messages to local path ${eventFile}`)
+      logger.info(
+        `Emitting ${uniqueTerms.length} ${termType} browse terms across ${bulkBrowseTerms.length} messages to local path ${eventFile}`
```
charmingduchess, line 60: the rest of the diff in this file is just formatting, but the following lines are additions

nonword (dismissed):

lib/browse-terms/index.js
```diff
@@ -51,18 +56,34 @@ const emitBrowseTerms = async (esDocuments, termType) => {
...
+      )
+      if (!fs.existsSync(process.env.BROWSE_TERM_OUTPUT_DIRECTORY)) {
+        fs.mkdirSync(process.env.BROWSE_TERM_OUTPUT_DIRECTORY, logger.error)
+      }
```
nonword: Minor point: You might move this `existsSync` call outside the loop so you're not calling it on every pass as it's a little expensive.

PR #348, opened 2026-08-05
Add option to bulk-index to write to csv

danamansana:
Adds an option to bulk-index script to write to csv instead of writing to ES.
Want to do some more testing on this but also want to get thumbs up on this approach, since it is a little different from what the ticket says. The other possibility is to write this as a separate script and do some refactoring of common elements into a utility, but this issue with that is that there is more in common than not with the existing script. Let me know what you think

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ sam local invoke --profile nypl-digital-dev -t sam.local.yml -e test/sample-even
 
 The bulk-indexer can be used to index lots of records at once by direct BibService SQL query.
 
-For example, to reindex the first 10K NYPL bibs with marc 001 in QA:
+For example, to reindex the first 10K NYPL bibs in QA:
 ```
 node scripts/bulk-index.js --envfile config/qa-bulk-index.env --type bib --hasMarc 001 --nyplSource sierra-nypl --limit 10000
 ```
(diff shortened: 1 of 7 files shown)


charmingduchess (changes requested):
I think it's important to separate these processes. I don't think we need to worry about maintaining the older querying logic. My current workflow, and the future of this workflow as I understand it assumes a csv input for the bulk index script. I'd be in favor of junking  all the SQL biz that isn't about this new csv generation flow. If that is spooky, we could also just extract the SQL logic to another file to clean this file up, and that could still expose the csv builder as a separately runnable script. Interested to hear @nonword 's take too.

scripts/identify-ids-by-marc.js
nonword, line 1: Small suggestion that we name this `identify-ids-by-marc` to match the other CSV generating scripts with a similar contract

scripts/identify-ids-by-marc.js
```diff
@@ -0,0 +1,181 @@
...
+    }
+    return new Pool(config)
+  },
+  connect: (name) => db.dbConnectionPools[name].connect(),
+  endPools: () => {
+    return Promise.all(
+      Object.values(db.dbConnectionPools).map((pool) => pool.end())
+    )
+  }
+}
```
nonword, line 67: Wondering if this common connection stuff shouldn't live in a common util, but not sure best place..

nonword (dismissed):

scripts/identify-ids-by-marc.js
danamansana: Ok I've renamed that

charmingduchess (approved):
Sorry for delay - looks good to me.

PR #349, opened 2026-08-05
&ai update bulk index script to write status to csv

danamansana:
- Adds some logic to write status to a file when bulk updating from a csv
- Can stop script and will restart from the same place
- Adds some tests

### DIFF
--- a/scripts/bulk-index.js
+++ b/scripts/bulk-index.js
@@ -148,6 +148,7 @@ const {
 } = require('../lib/prefilter')
 const {
   batch,
+  CsvProgress,
   groupIdentifierEntitiesByTypeAndNyplSource,
   delay,
   die,
@@ -682,6 +683,19 @@ const updateByCsv = async (options = { offset: 0 }) => {
     throw new Error('--csvIdColumn is required')
   }
 
+  const progress = await CsvProgress.forCsv(options.csv)
+
+  if (['completed', 'failed'].includes(progress.status())) {
+    logger.info(`Skipping CSV ${options.csv} because status is ${progress.status()}`)
+    return
+  }
+
+  if (progress.status() === 'running') {
+    options.offset = progress.offset
+  } else if (progress.status() === 'preparing') {
+    progress.updateOffset(options.offset)
+  }
+
   const rawContent = fs.readFileSync(options.csv, 'utf8')
   const rows = csvParse(rawContent)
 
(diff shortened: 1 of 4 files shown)


scripts/utils.js
```diff
@@ -401,6 +401,65 @@ class Timer {
...
+    }
+  }
+
+  static async forCsv (filepath) {
+    const progress = new CsvProgress(filepath)
+    if (fs.existsSync(progress.statusFilepath)) {
+      progress.state = JSON.parse(fs.readFileSync(progress.statusFilepath, 'utf8'))
+    } else {
+      try {
+        progress.state.count = await lineCount(filepath)
```
nonword, line 424: This addition worked so well that it highlighted a bug in existing code: `lineCount` uses `wc -l` to get line count, but CSVs often (arguably incorrectly) don't terminate in a newline, causing the strict _linecount_ to be one fewer than the actual number of records, which is going to cause this status file count to be off by one frequently. Can you change that to use `awk 'END {print NR}' filename.csv` to efficiently retrieve csv linecounts? (or just ticket it)

nonword (dismissed):
Looks great

scripts/utils.js
danamansana: Updated!

scripts/bulk-index.js
```diff
@@ -676,48 +677,87 @@ const castRowToIdentifier = (row, options) => {
...
+    throw e
+  }
 
   logger.info(`Processing ${options.csv} rows ${options.offset} to ${end} (${rowsToProcess.length} rows)`)
 
   // Test first row to determine whether we need to interpret as prefixed
   // identifiers or just plain numeric ids:
-  const identifiersHaveNyplSource = rowsToProcess[0].nyplSource
-  const identifiersHaveType = rowsToProcess[0].type
+  if (rowsToProcess.length > 0) {
```
charmingduchess: I've seen this nyplsource check be pretty finicky. Can you extract into a method and do some unit tests on a couple different csv configurations (prefixed ids, non-prefixed, w/ and w/o nypl source).

scripts/utils.js
```diff
@@ -401,6 +401,65 @@ class Timer {
...
+
+  status () {
+    return this.state.status
+  }
+
+  get offset () {
+    return this.state.offset
+  }
+
+  updateStatus (status) {
```
charmingduchess, line 446: Could be useful to have the statuses saved as constants, eg
```
const COMPLETED = 'completed'
updateStatus(COMPLETED)
```
 and/or add input validation here.

charmingduchess (changes requested):

scripts/bulk-index.js
danamansana: OK I've added some tests

scripts/utils.js
danamansana: Good idea, I've changed those to constants

scripts/bulk-index.js
```diff
@@ -702,38 +744,7 @@ const updateByCsv = async (options = {}) => {
...
-    } else if (!options.type && !identifiersHaveType) {
-      const errorMsg = 'Must specify --type if not apparent from CSV'
-      progress.addMessage(errorMsg)
-      progress.updateStatus('failed')
-      throw new Error(errorMsg)
-    }
-  }
+  const rowsToProcess = extractAndValidateIdentifiers(rows, options, progress, sourceMapper)
 
   if (progress.status() === 'preparing') {
```
charmingduchess: let's include preparing and running as constants too

charmingduchess (changes requested):

scripts/bulk-index.js
danamansana: Good catch, added!

charmingduchess (approved):

